PR #36, opened 2014-09-09
Adding HST+JWST band passes

srodney:
I've added .dat fiels with transmission functions into data/bandpasses/hst and /jwst 
for HST-ACS, HST-WFC3IR, HST-WFC3UVIS, JWST-NIRCAM. 

For JWST-MIRI we only have idealized tophat functions for the band passes, so no .dat files are needed.


kbarbary:
superceded by #57

PR #54, opened 2014-12-27
Python 3 already!

kbarbary:
This (finally) implements Python 3 support, using the six module. Currently, all tests pass on Py3 and Py2. However, I'm waiting on two things before merging this:

- iminuit (required for `fit_lc`) doesn't yet support python 3. Currently the `fit_lc` test is simply skipped if iminuit isn't installed. I'd like to get iminuit working on Py3 and ensure that fit_lc works.

- sncosmo still has only about 50% test coverage, so I'm worried that some untested things still don't work. Gotta up that test coverage first...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,5 +116,5 @@ setup(name=PACKAGENAME,
           'Topic :: Scientific/Engineering :: Physics'],
       cmdclass=cmdclassd,
       zip_safe=False,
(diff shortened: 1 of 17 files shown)


kbarbary:
Testing is up, at least a bit, and I'm getting tired of waiting on iminuit, so I'm merging this now.

PR #57, opened 2015-01-09
        Added HST and JWST band passes

srodney:
* HST : added .dat files for acs and wfc3 bands (plus some nicmos and wfpc2)
         * JWST : added preliminary bands for nircam
         * dirt simple test function added to test_registry


kbarbary:
Awesome, looks great!  The test failure is only due to the style checker that gets run automatically. It's pretty easy to fix to conform to the style guidelines using the pep8 checker, but I can do it after merging if you don't want to bother about it.

srodney:
Clearly I'm still getting the hang of pep8.  I think I got it with 5e6284c.

kbarbary:
I love how the docs automatically build these plots: http://sncosmo.readthedocs.org/en/latest/builtins/bandpasses.html but they don't look as good as they used to with the large wavelength range and number of filters of JWST. I'll see if I can do something about that.

kbarbary:
@srodney Do you recall how the files for WFC3 UVIS were generated? From  http://www.stsci.edu/hst/wfc3/ins_performance/throughputs/Throughput_Tables it looks like there are two separate transmissions for the two CCDS in the UVIS detector. Are the trasmissions here an average, or just one or the other CCD?

srodney:
I did a quick check with the F218W filter, and it confirmed my vague recollection:   I used the UVIS2 filter curves.

kbarbary:
Thanks for checking!

Issue #21, opened 2014-03-27
More robust wavelength bounds: minwave(), maxwave() in Model


kbarbary:
`Model.minwave()` and `Model.maxwave()` are currently determined based only on the `Source` instance of the model. They should also take into account the wavelength bounds of the `PropagationEffect` instances.

PR #58, opened 2015-01-13
Small changes in fitting.py and models.py

srodney:
Added checkeffects to bandoverlap and propagating min chi2 parameter values into best-fit model objects.

bandoverlap_checkeffects_test_ipynb

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -114,9 +114,9 @@ def flatten_result(res):
 def cut_bands(data, model, z_bounds=None):
 
     if z_bounds is None:
-        valid = model.bandoverlap(data['band'], checkeffects=True)
+        valid = model.bandoverlap(data['band'])
     else:
-        valid = model.bandoverlap(data['band'], z=z_bounds, checkeffects=True)
+        valid = model.bandoverlap(data['band'], z=z_bounds)
         valid = np.all(valid, axis=1)
 
     if not np.all(valid):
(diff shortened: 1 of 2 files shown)


kbarbary:
Bug fix in `fitting.py` looks good. I think we never noticed this omission because the chisq function sets the model parameters as the minimization runs. So at the end of the minimization, the model parameters end up being pretty close to the optimal values. I guess you found cases where the model parameters were close (but not identical) to those reported in the result object?

The intent with `model.bandoverlap` was to always check effects (see #21) so I think the default behavior is just a bug. I'll take a closer look to make sure.

kbarbary:
With `Model.bandflux` I think it would be sufficient to change `self._source.minwave() * (1. + z)` to `self.minwave()` and similarly for `maxwave()`.

kbarbary:
I meant `Model.bandoverlap` not `Model.bandflux` in the previous comment (now edited).

srodney:
"I guess you found cases where the model parameters were close (but not identical) to those reported in the result object?"

Yes... though I don't have a good test-case ready at hand.  

"With Model.bandoverlap I think it would be sufficient to change self._source.minwave() * (1. + z) to self.minwave() and similarly for maxwave()."

Yep.  I think you're right.  Fixed in the latest commit.

kbarbary:
Looks good.

As a side note, instead of doing `git merge upstream/master` to merge the latest commits to master into your branch, you can do `git rebase upstream/master`. The two methods are pretty similar in that you will have to resolve any conflicts that git cannot automatically resolve. The difference is that `git rebase` "rewrites" your commits so it looks like you started from where `upstream/master` is. The benefit being that the history is more linear and there aren't a lot of extra merge commits.

PR #62, opened 2015-01-18
Tophat bands

kbarbary:
@srodney should have a quick look at this.

This does two things for the JWST bandpasses:

1. Registers a loader instead of the actual bandpass. The primary benefit of this is that the bandpasses show up in the docs (the machinery for that is built around the loaders).

2. Changes the sampling of the bandpasses so that they are only sampled in their region of support, and sampled more finely (100 samples per bandpass). There's no reason to have the zero transmission points outside the region of support, so that's a good thing. However, I'm not really sure what that such a high sampling in the region of support is necessary as spectral models are likely to be pretty smooth at these wavelengths. Any input on this?

### DIFF
--- a/sncosmo/_builtin/bandpasses.py
+++ b/sncosmo/_builtin/bandpasses.py
@@ -337,9 +337,11 @@ def tophat_bandpass(ctr, width, name=None):
     """Create a tophat Bandpass centered at `ctr` with width `width` (both
     in microns. Sampling is fixed at 100 A == 0.01 microns"""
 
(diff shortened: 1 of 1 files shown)


srodney:
Looks good to me.   

re item 2: Spectral models will not be so smooth at mid-IR wavelengths when they get redshifted to z=5-15, which is one of the primary use cases for JWST SN science.   So we should keep them well sampled (even though the tophats are clearly just a crude approximation for now).

PR #63, opened 2015-01-20
Adding new CC and PopIII SN models to sources.py

srodney:
I tried to bring this in with a cleaner commit history by using `git rebase` instead of `git merge` ... but I'm not sure that I succeeded.   If this PR would result in duplications of previous commits, then let me know, and I'll rewind and rewrite history to clean it up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 # Ignore .c files by default to avoid including generated code. If you want to
 # add a non-generated .c extension, use `git add -f filename.c`.
 *.c
+*.so
 
 # Packages
 *.egg
@@ -47,3 +48,6 @@ sncosmo/sncosmo.cfg
 
 # pycharm files
 .idea
+
+# ipython notebooks
+*.ipynb
--- a/sncosmo/_builtin/sources.py
+++ b/sncosmo/_builtin/sources.py
@@ -332,6 +332,86 @@ registry.register_loader(Source, '2011fe', load_2011fe, version='1.0',
                                'url': 'http://snfactory.lbl.gov/snf/data',
                                'reference': p13ref})
 
+
+# --------------------------------------------------------------------------
+# SNANA CC SN models
+
+baseurl = 'http://sncosmo.github.io/data/models/snana/'
+meta = {'url': 'http://das.sdss2.org/ge/sample/sdsssn/SNANA-PUBLIC/',
+        'type': 'Ic',
+        'subclass': '`~sncosmo.TimeSeriesSource`',
+        'reference': ('SNANA', 'Kessler et al. 2009 '
+                      '<http://adsabs.harvard.edu/abs/2009PASP..121.1028K>'),
+        'note': "extracted from SNANA's SNDATA_ROOT on 5 August 2014."}
+
+for modelname, sedfile, type in [
+    ['Ic.01', 'CSP-2004fe.SED',  'Ic'],
+    ['Ic.02', 'CSP-2004gq.SED',  'Ic'],
+    ['Ic.03', 'SDSS-004012.SED', 'Ic'],
+    ['Ic.04', 'SDSS-013195.SED', 'Ic'],  # PSNID
+    ['Ic.05', 'SDSS-014475.SED', 'Ic'],
+    ['Ic.06', 'SDSS-015475.SED', 'Ic'],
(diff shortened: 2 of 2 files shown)


srodney:
And now I'm seeing the Travis CI build is failing... but I don't understand why.   The only change I wanted to get in was a fairly trivial update to sources.py, so maybe its not worth the trouble to clean this up, and I should just close this PR and try again.

srodney:
Yes, I see now that my rebase was borked.  I'm not quite sure what I did, but my branch has diverged too much, so I'm closing this and I'll try to backtrack.

kbarbary:
For the record, I would have expected rebase to get rid of the first 4 commits here, since they were already merged into master. Are you sure you `git fetch upstream`'d before rebasing? My other thought is that there might have been a problem with how you resolved conflicts. But not sure what that would be.

kbarbary:
On second look, I might not have expected that. There are a bunch of merge commits back and forth and  you're right that those 4 commits were already rewritten at some point. A true git master could probably salvage this, but honestly I'd probably just start from the current upstream/master and do it manually. (I don't think reset --hard would help here.)

once you get this figured out, you can `push -f` to srodney/master and this PR should be updated, so I'm reopening it.

srodney:
Alright, I think this is cleaned up now.  I did : 

```
git fetch upstream master   
git reset --hard ac09d56    # revert to last clean merge point
git merge upstream/master  # fast-forward to current upstream/master HEAD
# < updates to sources.py and .gitignore >
git commit
git push -f
```

kbarbary:
Looks good. The test failure is spurious.

How much do you like those model names? For the existing Sako models, I called them by their IAU names. Could we do the same here? They're certainly harder to remember, but we could add some convenience functions to get a list of model names. Something like

```python
>>> sncosmo.model_names('snana')
['2004fe', '2004gg', ... ]
>>> sncosmo.model_names('snana', 'Ic')
# return just the SNANA Ic model names...
```

(or maybe the model names should be `'snana-2004fe'` to be really explicit?) If we had this functionality, it wouldn't matter much how long the model names were because you'd never have to explicitly type them in.

Just a thought to make things more consistent.

srodney:
I'm not at all wedded to the short-cut model names.   The SNANA models are all the result of warping the same sub-class spectral templates to match a given light curve from SDSS, SNLS, CSP, etc.   So I figured naming them explicitly after the light curve model is slightly misleading... since that SDSS object is not providing a complete template.   But that's not really a serious concern.  

So, yes, flexibility of naming is fine with me.  I could work on that later this week, if you don't get to it first.

kbarbary:
Ah I see the logic in that. I'm going to be redoing the data-fetching infrastructure part soon anyway, so I'll go ahead and merge this now, and most likely change the names later. I'll think a bit more about what names would be best in conjunction with the infrastructure work.

BTW, are the Nugent spectral templates the ones used to create the SNANA models?

rbiswas4:
On Tue, Jan 20, 2015 at 2:07 PM, Kyle Barbary <notifications@github.com>
wrote:

> Ah I see the logic in that. I'm going to be redoing the data-fetching
> infrastructure part soon anyway, so I'll go ahead and merge this now, and
> most likely change the names later. I'll think a bit more about what names
> would be best in conjunction with the infrastructure work.
>
> BTW, are the Nugent spectral templates the ones used to create the SNANA
> models?
>
> Yes. It is Nugent + warping prescription based on light curves of a
supernova. So,  tacking on 'snana' as part of the model  to indicate the
prescription would probably be a good idea.

kbarbary:
Have a look at the new names:
http://sncosmo.readthedocs.org/en/latest/builtins/sources.html

I like the `snana-` prefixes on the SNANA models because it gives some indication of the procedure used to create the model (e.g., starting from the Nugent models, etc). The `snana-sdss000018` names are a bit verbose though.  We could change them to `snana-000018` but that seems obscure. I'd rather look up their IAU names and use something like `snana-2004hx`. It makes it obvious that the second part refers to a specific SN.

srodney:
The new names look good to me.  I've just issued [PR #68](https://github.com/sncosmo/sncosmo/issues/68)  and an accompanying [PR #2 in sncosmo.github.io](https://github.com/sncosmo/sncosmo.github.io/pull/2) which updates to IAU SN names for SDSS sources whenever possible.

PR #65, opened 2015-01-22
Plot lc doc

rbiswas4:
Tried to modify the documentation for photometric data tables. Linking seems to still be a problem

### DIFF
--- a/docs/photdata.rst
+++ b/docs/photdata.rst
@@ -26,7 +26,9 @@ looks like, you can load an example with the following function::
     ... etc ...
 
 Additionally, metadata about the photometric data can be stored with
-the table: ``data.meta`` is an `OrderedDict` of the metadata.
+the table: ``data.meta`` is an `OrderedDict` of the metadata. Additional
+columns in the table can exist, but the above columns (with aliases for
+the names defined in the next section) are necessary.
 
 
 Column names and units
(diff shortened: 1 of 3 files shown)


kbarbary:
You have the following forms:
```
 `~doc.photdata`  # in fitting.py
:doc:`~photdata`  # in fitting.py
:doc:`photdata`  # in plotting.py
```

I would expect only the last one to work. Does it?

By default, the backticks reference a code object. so `sncosmo.fit_lc` in backticks creates a link to the documentation on that object. The tilde `~` (e.g., `~sncosmo.fit_lc`) makes it so the link is still right, but only `fit_lc` is displayed in the docs: the module prefix is left off. Prepending `:doc:` in front of the backticks changes the "role": it tells markdown that you're referencing a rst document, rather than an object. That's why the first two forms above shouldn't work: photdata is an rst document, not an object in sncosmo itself.

rbiswas4:
The last one was the one I started with. It did not work on my computer. I was trying different versions (without knowing the syntax logic) but did not find one that worked. I was wondering if it has something to do with my setup and whether things would be different on rtd. Hence the PR.

kbarbary:
You seem to be right: the normal
```
:doc:`photodata`
```
doesn't work in docstrings. I just wrote "see the 'photometric data' section of the documentation" (no hyperlink).

rbiswas4:
Ok ... I am deleting the branch with hanging commits.

PR #68, opened 2015-01-28
updated snana sources with IAU names

srodney:
and downcased the whalen source names in the registry loader.

See also the [pull request](https://github.com/sncosmo/sncosmo.github.io/pull/2) in sncosmo.github.io to update the urls.json file accordingly:


kbarbary:
Looks good to me. I'll wait until we merge sncosmo.github.io#2 first though.

kbarbary:
BTW, sorry I've been shifting things around in builtins.py so much recently. Glad to see you could still find your way around.

I realized recently that with how I've set it up now, the `urls.json` file is fetched once per session, regardless of whether or now the data has already been downloaded. This is undesirable for cases where you want to work offline. I'm working on a fix but it will take a bit more time.

PR #71, opened 2015-02-05
bug fix in Model.bandoverlap()

srodney:
The bug was introduced in #58 when I tried to simplify the bandoverlap bugfix. 

The simplification resulted in a double-counting of the redshift in cases where the model had an existing redshift value set.  With this PR, if z is not None, then we remove the model redshift before shifting to the user-specified value of z to check for overlap.

### DIFF
--- a/sncosmo/models.py
+++ b/sncosmo/models.py
@@ -1231,7 +1231,6 @@ class Model(_ModelBase):
         overlap : bool or `~numpy.ndarray`
 
         """
-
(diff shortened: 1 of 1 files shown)


kbarbary:
Looks good. Would be good to have a test for this, to avoid future problems. Could go after `test_minwave` and `test_maxwave`: https://github.com/sncosmo/sncosmo/blob/master/sncosmo/tests/test_models.py#L136

Issue #69, opened 2015-01-28
should Model.flux() return zero at t<mintime ?


srodney:
Some of the Whalen et al. Pop III CC SN models include a sharp peak in flux in the very first defined epoch (the shock breakout peak).  These tend to induce a bad Spline2D extrapolation to times preceding that first defined epoch.  For example: 

![popiiisn](https://cloud.githubusercontent.com/assets/1355473/5938594/7fbbe956-a6ca-11e4-93f6-d9039dd6e15f.png)

Physically, the flux in the circled regions should be zero (and magnitudes should be undefined or extremely large) right up to the shock breakout point, which is given by mintime().    

Should the flux() function use a masked array to flag epochs that precede mintime?
Or should it just return zeros?   

Perhaps a new model parameter is needed to define the behavior before mintime?

Should the behavior of flux be adjusted similarly for times after maxtime? (there is less danger of bad extraopolations at late times)

kbarbary:
Yes, `Model.flux()` should indeed return zero before mintime. Unfortunately, there's no option in the `RectBivariateSpline` class to make this happen automatically; for points outside the "bounding box" of the spline, it simply returns the nearest point. For most (all?) other models, this hasn't been an issue because the first phase of the model is at zero flux, so the model does return zero before mintime.

Returning zero flux (rather than a masked array) is the desired behavior: The flux of the model before mintime is not *unknown* (as a masked array would imply). The flux of the model is zero (or close enough to zero that we don't care). Having a well defined flux before mintime make it possible to fit for `t0`: When doing such a fit, the model can shift in time so that the early data points might be either before mintime or after mintime, and we wish to always be able to compare the data points to the model.

The behavior after maxtime is a bit trickier because the models don't usually extend to the point where the flux is close enough to zero that we don't care. Related: #7.

If possible, the easiest thing would be to add a point at zero flux to the Whalen models, slightly before the first epoch. This might induce some crazy spline ringing though. I wish we could add the option directly to `RectBivariateSpline`, but the out of bounds behavior is implemented in the underlying Fortran code; we'd have to PR to scipy and wait for the change to propagate to a release. An alternative is to subclass `RectBivariateSpline` to explicitly check for points before mintime and use this subclass for the Whelan models.

Can you do a quick check to see if adding a point at zero flux just before the first epoch does indeed induce spline ringing? If so, we can go the subclassing route.

srodney:
[...]

PR #72, opened 2015-02-08
Adding phase<minphase checks to _flux() 

srodney:
See issue #69


kbarbary:
Looks like there's a real test failure this time.
```
sncosmo/tests/test_models.py:25: TestTimeSeriesSource.test_flux FAILED
```
Care to investigate?

Also, can you take out the changes for SALT2? The model is basically zero at the first phase, so the extra check isn't really necessary. If we want it to be identically zero, we can alter the first phase to be all zeros when we initialize the model.

kbarbary:
Superseded by #77.

PR #77, opened 2015-02-17
add 'zero_before' keyword for Whalen models

kbarbary:
This is my solution to #69, an alternative solution to #72.

It has the benefit of being minimally intrusive to models that don't need the check. (In fact, you don't want the check for most models because it creates a discontinuity that will make fitting harder.) There's a lot of different ways you could accomplish the same thing. For example, one could implement a subclass of `RectBivariateSpline` or of `TimeSeriesSource` that does the check. This solution simply adds a keyword argument to the initializer in `TimeSeriesSource`. This seems like the simplest thing to do for now.

With this change:
```python
In [5]: model = sncosmo.Model(source='whalen-z15b')

In [6]: model.source.minphase()
Out[6]: 1.6599999999999999

In [8]: model.flux(1.6, [100., 200.])
Out[8]: array([ 0.,  0.])
```
Does this look good @srodney?

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -137,10 +137,11 @@ def tophat_bandpass(ctr, width, name=None):
     return Bandpass(wave, trans, wave_unit=u.micron, name=name)
 
 
-def load_timeseries_ascii(relpath, name=None, version=None):
+def load_timeseries_ascii(relpath, zero_before=False, name=None, version=None):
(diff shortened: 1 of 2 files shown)


srodney:
This looks good to me, and saves me from having to clean and debug #72.

As far as I'm concerned, you can merge this, then close #72 and issue #69.

PR #80, opened 2015-03-01
Added warning when one supplies guess_amplitude_bound=true but

danzimm:
doesn't include any amplitude parameter to vary.

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -714,6 +714,9 @@ def nest_lc(data, model, vparam_names, bounds, guess_amplitude_bound=False,
     data = cut_bands(data, model, z_bounds=bounds.get('z', None))
 
     if guess_amplitude_bound:
+        if 'x0' not in vparam_names and 'amplitude' not in vparam_names:
+            warn("Trying to guess the amplitude bounds but neither 'amplitude'"
+                 "nor 'x0' are in vparam_names.", RuntimeWarning)
         if model.param_names[2] in bounds:
             raise ValueError("cannot supply bounds for parameter {0!r}"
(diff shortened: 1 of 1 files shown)


kbarbary:
You can use `model.param_names[2]` instead of `'x0'` or `'amplitude'`. That is, `model.param_names[2]` will be `'x0'` for the SALT2 model and `'amplitude'` for other models. This is better so that the specific name of the amplitude parameter isn't hard-coded.

Looking at the code, I think we should raise a `ValueError` instead of just a warning, as this is almost certainly due to user error. If we just raise a warning, then the amplitude bound still gets added to `bounds`, which is weird when it isn't even being varied!

kbarbary:
PS, I just noticed that the next ValueError:

```python
ValueError("cannot supply bounds for parameter {0!r}"
           " when guess_amplitude_bound=True")
```

is missing a format argument. It should be

```python
ValueError("cannot supply bounds for parameter {0!r}"
           " when guess_amplitude_bound=True"
           .format(model.param_names[2]))
```

You can write something similar for the guess_amplitude_bound error message.

kbarbary:
PS @danzimm You can add a new commit to your branch and push it again and the changes will be reflected here. Let me know if you can't make the suggested changes and I'll try to do it instead.

danzimm:
I can definitely do it - sorry was busy and didn't flag this the first time it came by. Flagging now!

kbarbary:
@danzimm Thanks! (no rush, just wanted to make sure this doesn't get forgotten.) BTW, there's been a new commit on master, so you might want to `git rebase upstream/master` on your branch.

PR #84, opened 2015-04-04
builtins: added bytes-to-string decoder since urllib.urlopen() returns bytes instead of a string in Python 3

### LINKED ISSUE #83: urllib.urlload() and json.load() break when importing a SN model in Python 3 for the first time
I tried importing the built-in supernova model 'hsiao' following the instructions on the SNCosmo website under the "Supernova Models" section. In Python 3, the line

`model = sncosmo.Model(source='hsiao')`

breaks with the error `TypeError: can't use a string pattern on a bytes-like object`. The error comes from this line in `builtins.py`:

`urls = json.load(f)`

which fails because the previous line,

`f = urlopen("http://sncosmo.github.io/data/urls.json")`

returns raw bytes in Python 3, whereas in Python 2 it returns a string (which is what `json.load()` expects).

This error does not occur in either version of Python if the requested SN model has already been downloaded and cached.

bcfriesen:
This fixes a bug when trying to import a supernova model for the first
time in Python 3. json.load() needs a string argument, but in Python 3,
urllib.urlopen() returns raw bytes. (In Python 2 it returns a string as
the code expects currently.) The fix yields identical behavior in both
Python 2 and 3.

Note that the bug does not trigger in either version of Python if the
requested model has already been downloaded and cached.

This fixes #83.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -11,6 +11,7 @@ import tarfile
 import warnings
 import os
 from os.path import join
+import codecs
 
 import numpy as np
 from astropy import wcs, units as u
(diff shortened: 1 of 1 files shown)


kbarbary:
It took me a fair bit of googling to fully understand the issue, but your solution seems to have nailed it. Thanks!

(PS: Its nice to have someone actually running Python 3 to catch this sort of thing. I've avoided remote data in the test suite, so this never got tested there.)

PR #86, opened 2015-04-07
io: force bytes-to-string decoding of "line" argument to _stripcomment()

### LINKED ISSUE #85: read_griddata_ascii() passes raw bytes to _stripcomment() in Python 3
When downloading some supernova models, e.g., 'nugent-sn1a' (but not all, e.g., 'hsiao'), read_griddata_ascii() is invoked. This function passes each line of an ASCII file to _stripcomment(). However in Python 3 the line object is interpreted as raw bytes, while in Python 2 it is a string, which is what _stripcomment() expects. As a result, Python 3 throws the TypeError "Type str doesn't support the buffer API".

bcfriesen:
When importing some of the built-in supernova models for the first time
(e.g., `s11-2006jl`), the model data is downloaded and parsed with
`read_griddata_ascii()`. In `read_griddata_ascii()` each line from the file
object is passed to `_stripcomment()` to remove comment text. However in
Python 3 the `line` argument to `_stripcomment()` is raw bytes, which
leads to the TypeError `Type str doesn't support the buffer API`. In
Python 2 the error does not trigger because the `line` argument is
interpreted as a string.

This fixes #85 by explicitly decoding the `line` argument as a string.

### DIFF
--- a/sncosmo/io.py
+++ b/sncosmo/io.py
@@ -74,7 +74,7 @@ def read_griddata_ascii(name_or_obj):
     x1_current = []
     y1_current = []
     for line in f:
-        stripped_line = _stripcomment(line)
+        stripped_line = _stripcomment(line.decode("utf-8"))
         if len(stripped_line) == 0:
             continue
         x0_tmp, x1_tmp, y_tmp = map(float, stripped_line.split())


kbarbary:
Would it be enough to just open the file in text-mode? On line 65 of io.py, the file is being opened in binary mode, for some reason:

```python
        f = open(name_or_obj, 'rb')  # should just be 'r'
```
It looks like there are some other places in io.py where binary mode should be removed as well. 

There is a unit test for this function but it uses `StringIO` for input so it never caught this!

bcfriesen:
Yeah, thought about that too, but changing 'rb' to 'r' won't fix the problem if that "else" is triggered on line 66.

If there are other places with this stupid bytes-vs-string business, I haven't hit them yet.

kbarbary:
I think its OK to require that if `name_or_obj` is a file object (if `else` is triggered on L66), it should be a "text file" (a file opened in text mode). For example, that's what the json module requires:

```python
>>> f = open('tmp.txt', 'wb')
>>> json.dump([1, "simple", "list"], f)
  File "/usr/lib/python3.4/json/__init__.py", line 179, in dump
    fp.write(chunk)
TypeError: 'str' does not support the buffer interface
>>> 
```

Internally, sncosmo doesn't pass file objects to `read_griddata_ascii` anywhere (except in tests), so no other internal changes are necessary.

bcfriesen:
Ah OK. Then yes, I agree that changing "rb" to "r" on line 65 is the better choice. In that case, ignore this PR and I'll send another one with that change.

kbarbary:
Awesome, thanks!

PR #88, opened 2015-04-09
Model: add color() member function

bcfriesen:
This returns the difference in magnitudes between two filters. Even
though it simply invokes `Model.bandmag()` twice, we require that the two
bandpass arguments be scalars, since lists of pairs (or pairs of lists)
of filters or magnitude systems leads to cumbersome output. The only
allowable iterable is the `phase` argument.

### DIFF
--- a/sncosmo/models.py
+++ b/sncosmo/models.py
@@ -13,6 +13,7 @@ from scipy.interpolate import (InterpolatedUnivariateSpline as Spline1d,
                                RectBivariateSpline as Spline2d,
                                splmake, spleval)
 from astropy.utils import OrderedDict as odict
+from astropy.utils.misc import isiterable
 from astropy import (cosmology, units as u, constants as const)
 from astropy.extern import six
 
@@ -1386,6 +1387,42 @@ class Model(_ModelBase):
             The return value is an `~numpy.ndarray` if any are interable.
         """
         return _bandmag(self, band, magsys, time)
+
+    def color(self, band1, band2, magsys, phase):
+        """Color at the given time(s) through the given pair of bandpasses, and
+        for the given magnitude system.
+
+        Parameters
+        ----------
+        band1 : str
+            Name of first bandpass in registry.
+        band2 : str
+            Name of second bandpass in registry.
+        magsys : str
+            Name of `~sncosmo.MagSystem` in registry.
+        time : float or list_like
+            Observer-frame time(s) in days.
+
+        Returns
+        -------
+        mag : float or `~numpy.ndarray`
+            Color for each item in time, band, magsys.
+            The return value is a float if all parameters are not iterables.
+            The return value is an `~numpy.ndarray` if phase is iterable.
+        """
+
+        if (((isiterable(band1)) and
+           not (isinstance(band1, six.string_types))) or
+           ((isiterable(band2)) and
+           not (isinstance(band2, six.string_types)))):
+            raise TypeError("Band arguments must be scalars.")
+
+        if ((isiterable(magsys)) and
+           not (isinstance(magsys, six.string_types))):
+            raise TypeError("Magnitude system argument must be scalar.")
+
+        return (self.bandmag(band1, magsys, phase) -
+                self.bandmag(band2, magsys, phase))
 
     def source_peakabsmag(self, band, magsys, cosmo=cosmology.WMAP9):
         return (self._source.peakmag(band, magsys) -


bcfriesen:
This is more an RFC than a PR, but GitHub doesn't provide a good mechanism for RFCs. Any interest in a function like this, Kyle? It's easy enough to invoke Model.bandmag() twice in user code rather than building it into the Model class, so I understand if you'd rather not bloat the code base with such a trivial function. But I for one would find it useful.

rbiswas4:
I used to have user-end code to do this, and was beginning to use  it in multiple cases. Another person seems to make it useful enough that this would be nice to have in sncosmo itself.

kbarbary:
Hmmm... I'm hesitant to add this. It might take longer to read the documentation and determine that the method does what you think it does and how to use it rather than just doing `m.bandmag(...) - m.bandmag(...)` yourself. Also, if we add it to `Model`, we would probably want to add it to `Source` for consistency.

Still, we might want to add it. I'll think about it some more and let this sit in the meantime.

By the way, PRs are often used for RFCs in this way, which I think is totally fine. On the bigger projects you often see PR titles prefixed with "[RFC]" or "RFC:" (also "WIP" for work in progress).

rcthomas:
I disagree about the argument about how long it takes to read the docs vs
use it.  I think "color" is pretty intuitive, and if we are going to make
the argument about time, then one time to read a doc on "color" and then
for the rest of your life be able to type "m.color()" instead of
"m.bandmag() - ..." will integrate out to a win for color.

This is a clear enhancement to the interface, not clutter, so I'd vote in
favor of it.  Whether the implementation is actually a full computation of
bandmag 2x that could be more optimized is something to worry about later.

Rollin

On Fri, Apr 10, 2015 at 9:08 AM, Kyle Barbary <notifications@github.com>
wrote:

> Hmmm... I'm hesitant to add this. It might take longer to read the
> documentation and determine that the method does what you think it does and
> how to use it rather than just doing m.bandmag(...) - m.bandmag(...)
> yourself. Also, if we add it to Model, we would probably want to add it
> to Source for consistency.
>
> Still, we might want to add it. I'll think about it some more and let this
> sit in the meantime.
>
> By the way, PRs are often used for RFCs in this way, which I think is
> totally fine. On the bigger projects you often see PR titles prefixed with
> "[RFC]" or "RFC:" (also "WIP" for work in progress).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/sncosmo/sncosmo/pull/88#issuecomment-91603518>.
>



-- 
Dr. R. C. Thomas
Computational Cosmology Center
Lawrence Berkeley National Lab
1 Cyclotron Road Mail Stop 50B4206
Berkeley, California 94720-8151
510.486.4697

kbarbary:
@rcthomas Fair point.

@bcfriesen Minor style point: change the return line so that it's less than 80 characters. E.g.,
```python
return (bandmag(...) -
        bandmag(...))
```

There's an automatic PEP8 style checker that gets run as part of the Travis tests (partially to keep me honest!). That's causing the test "failure" here.

The type checks won't catch numpy arrays (`isinstance(np.ndarray, list)` is False). I'll have to remind myself the best way to do this check, but it's probably using `from astropy.utils import isiterable`.

bcfriesen:
Thanks for the feedback. The 80-character wrap is fixed. Unfortunately, Python strings are iterables, which means the traditional tests for iteration will always throw false positives as long as you allow both strings and `Bandpass` objects to be used as arguments. (The same goes for `MagSystem` objects, right?) In light of that constraint, the new commit tests the `band{1,2}` and `magsys` arguments both for naive iterability (using `astropy.utils.misc.isiterable()`), as well as for being `six.string_types`. This should catch most cases. It's clunky but I don't know a better way to do it.

I `push --force`-ed the updated commit with the above issues fixed to bcfriesen/sncosmo@49defa8b99a68bc6f429d2dd773a4e58fede710f, which is still the tip of the `colors` branch, so this PR will still get the new changes.

Thoughts?

kbarbary:
Now I recall that I had to do the same explicit test for strings elsewhere. Annoying, but I don't know a better way.

pep8 is striking again I think. Probably something about unnecessary parentheses in conditionals and using `(` for line wrapping rather than `\`. You can check this locally by installing the pep8 package and doing `pep8 sncosmo` or `pep8 sncosmo/models.py` (from the root repo directory). 

If you find pep8 stuff too annoying, I can merge this and do it later. I know it can be annoying but it's the simplest solution I've found for ensuring a consistent style.

bcfriesen:
OK pep8 is happy this time. Latest iteration pushed to `colors` branch.

And style complaints are as valid as anything else. Keep fighting the good fight. I read LKML with my morning coffee and relish the subsystem maintainers tearing apart horribly formatted patches. There's a neat video on YT of Greg Kroah-Hartman explaining why formatting matters - because we're creatures of habit and we read code both faster and more correctly if it has a consistent pattern.

kbarbary:
> because we're creatures of habit and we read code both faster and more correctly if it has a consistent pattern.

:+1: :+1: 

Merging...

PR #89, opened 2015-04-12
typo fixed

rbiswas4:
typo in new color function: fixing phase to time to keep consistent with previous usage


kbarbary:
Looks good, but your line is >79 characters, triggering the test failure.

https://travis-ci.org/sncosmo/sncosmo/jobs/58201962#L621

rbiswas4:
fixed. Did not realize I had pushed the color direction in docstring ... I was a bit confused while using.

PR #93, opened 2015-05-17
Use astropy.extern.six for import in plotting.py

mwcraig:
I'm guessing that the import of range is intended to be from astropy's copy of six since six isn't listed as a dependency in setup. This PR corrects the import (and is motivated by work on auto-building conda packages for astropy-affiliated packages).

### DIFF
--- a/sncosmo/plotting.py
+++ b/sncosmo/plotting.py
@@ -7,7 +7,7 @@ import math
 import numpy as np
 from astropy.utils.misc import isiterable
 from astropy.extern import six
-from six.moves import range
+from astropy.extern.six.moves import range
 
 from .models import Source, Model, get_source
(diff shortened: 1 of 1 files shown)


kbarbary:
I must have thought that the `six` here would resolve to the `six` in the local scope (imported on the previous line `from astropy.extern import six`. Is that not the case? I guess this is clearer anyway.

Thanks for working on conda builds, Matt!

mwcraig:
It is not the case -- `from six.moves import range` looks for a six installed at top level, separate from the astropy bundled `six`.

Thought about changing all of the instances of `range` to `six.moves.range` (which would use the astropy `six`), but it looked like it wasn't a simple search/replace.

kbarbary:
Thanks, good to know. Merged.

PR #96, opened 2015-07-07
Fix Semantic Versioning URL

wmwv:
Fixes two backslashes that should have been forward slashes to fix semver.org reference.


kbarbary:
Thanks!

PR #106, opened 2015-08-25
Added scatter variable to realize_lcs so that we can turn off scatter …

rbiswas4:
…in realized light curves

### DIFF
--- a/sncosmo/simulation.py
+++ b/sncosmo/simulation.py
@@ -109,7 +109,7 @@ def zdist(zmin, zmax, time=365.25, area=1.,
 
 
 def realize_lcs(observations, model, params, thresh=None,
-                trim_observations=False):
+                trim_observations=False, scatter=True):
     """Realize data for a set of SNe given a set of observations.
 
     Parameters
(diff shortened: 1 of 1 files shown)


sncosmo/simulation.py
```diff
@@ -128,6 +128,11 @@ def realize_lcs(observations, model, params, thresh=None,
         If True, only observations with times between
         ``model.mintime()`` and ``model.maxtime()`` are included in
         result table for each SN. Default is False.
+    scatter : bool, optional, defaults to True
```
kbarbary: Nitpick: "Default is True" should go at the end of the description, not on this line. See above parameter for example.

rbiswas4:
Fixed docstring by moving the default description to the end. Is that just the convention in this code, or is it something closer to pep8 and I should be trying to follow that even more generally?

Could not figure out why is the travis ci check failing ?

kbarbary:
Change looks good. The specific format of docstrings are not specified by PEP8. Different projects use different conventions, but astropy and sncosmo use the numpydoc convention, which is detailed here: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt (search for "default"). The location of the default description is a very minor thing, but its better to just be consistent with the rest of the code.

The travis failure is caused by a PEP8 error (line too long). run `pep8 sncosmo --exclude=version.py` locally to reproduce. You can also see this on travis by clicking on "Details" in the box above, then clicking on the failing build, then reading the log.

rbiswas4:
> The location of the default description is a very minor thing, but its better to just be consistent with the rest of the code.

Completely agree on the consistency issue.

PR #112, opened 2015-09-30
bumped astropy to 0.4.0

ufeindt:
I tried installing on a machine with astropy 0.3.0 (the current version in the Ubuntu 14.04 repositories).

When I try importing I get the following error:
```
>>> import sncosmo
  File "/home/ufein/.virtualenvs/sncosmo-test/local/lib/python2.7/site-packages/sncosmo/__init__.py", line 124, in <module>
    from astropy.config import ConfigItem, ConfigNamespace
ImportError: cannot import name ConfigItem
```

`ConfigItem` does not exist before astropy 0.4.0, so I the requirement should be bumped to the version. I quickly verified that it work with that version in separate virtualenvs.

Unless I'm mistaken `pip-requirements` is all that needs to be changed.

### DIFF
--- a/pip-requirements
+++ b/pip-requirements
@@ -1,3 +1,3 @@
 numpy>=1.5.0
 scipy>=0.9.0
-astropy>=0.3.0
+astropy>=0.4.0


kbarbary:
Yep, you're right. The minimum astropy version should now be v0.4.0.

You should be able to install an older version of sncosmo (maybe v0.4?) which will work with astropy v0.3.

Thanks!

PR #115, opened 2015-11-07
Add CSP Bandpasses

dannygoldstein:
From [http://csp.obs.carnegiescience.edu/data/filters](http://csp.obs.carnegiescience.edu/data/filters).


kbarbary:
Sweet. A couple notes:

- All the existing built-in bandpass names are all lowercase. Can you change, e.g., `cspB` to `cspb` to follow this convention?

- Remove extra spaces after `,` in builtins.py to follow PEP8 and get tests passing.

- `csphs` / `csphd` for denoting Swope and Dupont is OK but seems pretty cryptic (even more so than band names normally are). On the other hand `csphswope` /`csphdupont` gets hard to read. I've boxed us in here by not using `_` in any existing bandpasses! If there's no good ideas we can stick with the extremely short names.

dannygoldstein:
First and third bullets are done. For the third, I added a column to `csp_filter_info.dat` that indicates the telescopes (Swope or DuPont) to which the filters are attached. The second bullet was OK as of #e7a62f. 

All unit tests are passing.

dannygoldstein:
Hold on. There is a typo on the CSP website. On http://csp.obs.carnegiescience.edu/data/filters, the second V band filter is called "V-3099", whereas in Stritzinger et al. (2011), it's called "V-3009". Updating this here.

dannygoldstein:
This is ready to go.

kbarbary:
OK, sounds good to me.

Issue #116, opened 2015-11-07
plot_lc for tables with mag and magerr?


dannygoldstein:
In my fork of `sncosmo`, I wrote a reader function `_read_csp` that reads photometry files released by CSP and returns tables like this:

```python
In [4]: sncosmo.read_lc('SN2006luopt+nir_photo.dat', format='csp')
Out[4]:
<Table masked=False length=134>
    mjd      filter   magerr   mag
  float64   string64 float64 float64
----------- -------- ------- -------
 54040.3211        u   0.026  18.793
 54040.3211        B   0.011  18.059
 54040.3211 V_LC9844    0.01  17.911
 54040.3211        g   0.007  17.919
 54040.3211        r   0.009  17.893
 54040.3211        i   0.017  18.473
54041.35904        Y   0.098  19.016
54042.32148        u   0.035  19.011
54042.32148        B    0.01  18.184
54042.32148 V_LC9844   0.011  17.987
54042.32148        g   0.009  18.031
54042.32148        r   0.012  17.968
54042.32148        i    0.02  18.583
54043.31674        Y   0.146  19.264
54043.31674        J   0.542  19.806
  54044.365        r   0.021  18.045
  54044.365        i   0.046   18.69
        ...      ...     ...     ...
54121.27083 V_LC9844   0.036  20.918
54121.27083        g   0.037  21.348
54121.27083        r   0.028  20.749
54121.27083        i   0.068  21.245
 54125.2998        B    0.06  21.739
 54125.2998 V_LC9844   0.042  20.956
 54125.2998        g   0.036  21.296
 54125.2998        r   0.035  20.886
 54125.2998        i   0.094  21.202
54130.21164        g   0.081  21.484
54130.21164        r   0.051  20.982
54130.21164        i   0.126   21.15
54131.24436 V_LC9844   0.065  21.029
54131.24436        r   0.056  21.033
54543.12044        Y    0.29  21.354
54838.24365        Y   0.245  21.474
54838.24365        J   0.435  21.792
54838.24365        H   0.741  19.715
```

I would like to plot these light curves using `plot_lc`, but this requires that the lcs be in flux space. Is there any deep reason for not allowing `plot_lc` to plot magnitudes?

kbarbary:
[...]

PR #117, opened 2015-11-09
add natural magnitude systems to sncosmo + more support for CSP data

dannygoldstein:
This PR augments the existing `MagSystem` class hierarchy in `sncosmo` by defining a `NaturalMagSystem` class. The purpose of this enhancement is to allow users to define magnitude systems in which the magnitudes / fluxes are calibrated such that they correspond to what one would measure through filters attached to a specific telescope or set of telescopes, rather than what one would measure in some standard set of filters or in a magnitude system based on a _single_ spectrophotometric standard. `NaturalMagSystem` inherits directly from `MagSystem`, and can be instantiated given the zeropoints, bands, and SEDs of the spectrophotometric standards of each band of the natural system.  Public interfaces to existing `MagSystems` (`SpectralMagSystem`, `ABMagSystem`) remain unchanged, but they now inherit from another abstract class, `StandardMagSystem`. This class is essentially what `MagSystem` was before. `MagSystem` has been stripped down somewhat, but its abstract interface has not changed.  

This PR also provides a new format for the `read_lc` method, `format='csp'`, as discussed in #116. 
-  [x] Make `read_lc(format='csp')` conform to the `sncosmo` data model

The CSP natural magnitude system (http://csp.obs.carnegiescience.edu/data/filters) is implemented as a builtin. 

Some unit tests are provided for the `NaturalMagSystem` class using the CSP natural photometric system.

Some of the commits in this PR are also present in #115.


kbarbary:
With #115 now merged, can you rebase this on master to make it easier to review?

dannygoldstein:
Done.

kbarbary:
Interestingly, I originally included this sort of functionality in sncosmo, calling it a "local" magnitude system. This [this removal commit](https://github.com/sncosmo/sncosmo/commit/0bc63b6a041dceafb4a9f5498de50a6a42e4c762) for what my minimal version looked like before it was removed. Having a separate class for this like you've done here seems like a better idea.

A couple incomplete thoughts:

- Terminology: as far as I know, in the context of magnitude systems, "natural" refers to the bandpass the magnitude is measured in, not the standard it is referenced to. For example, reporting Vega magnitudes in the desg, desr, ... filters from DES measurements would be natural magnitudes and also Vega magnitudes. On the other hand, applying some color corrections and reporting Bessell magnitudes from DES observations would not be natural magnitudes. This is why I chose the terminology "local" magnitude system when trying to implement this myself.

- Some such systems are referenced to AB, such as "SDSS magnitudes" which are AB, but with some fixed small offset in each band. To allow for such systems, it might be better for `NaturalMagSystem` to reference other system instances internally.

.gitignore
```diff
@@ -50,3 +50,4 @@ sncosmo/cython_version.py
 
 # ipython notebooks
 *.ipynb
+.cache/
```
kbarbary: What creates `.cache`?

sncosmo/spectral.py
```diff
@@ -468,16 +500,89 @@ def zpbandflux(self, band):
...
+    def bands(self):
+        return self._bands
+    
+    @property
+    def standards(self):
+        return self._standards
+
+    @_verify_band(0)
+    def _refspectrum_bandflux(self, band):
+        return self.standards[band].bandflux(band)
```
kbarbary: I think we could forgo the use of a decorator here, and just use a normal method. This is just as compact:

```python
def _refspectrum_bandflux(self, band):
    self._verify_band(band)
    return self.standards[band].bandflux(band)
```
and easier to parse.

sncosmo/spectral.py
```diff
@@ -468,16 +500,89 @@ def zpbandflux(self, band):
...
+    standards: iterable of `~sncosmo.Spectrum`,
+        The spectrophotmetric flux standards for each band.
+    """
+
+    def __init__(self, bands, zeropoints, standards, name=None):
+        super(NaturalMagSystem, self).__init__(name)
+        if len(bands) != len(zeropoints):
+            raise ValueError, 'Number of passed bands (%d) does not equal '  \
+                'the number of passed zeropoints (%d).'  % (len(bands),      
+                                                            len(zeropoints))
```
kbarbary: Minor style notes for consistency with rest of code base: (1) Instantiate exceptions with parentheses, (2) don't use `\` for line continuation, (3) use `.format` over `%`. E.g.,
```python
raise ValueError('Number of passed bands ({}) does not equal '
                 'the number of passed zeropoints({}).'
                 .format(len(bands), len(zeropoints)))

kbarbary:
Superceded by #126.

PR #125, opened 2015-11-21
implement radial bandpass

dannygoldstein:


kbarbary:
It would be helpful if you could remove the last commit from this PR. We're going to move to not including bandpasses alongside the source code. Instead, they'll be hosted externally, like models and spectra.

Probably the best way to do it is:
```
git checkout nonuniform
git branch megacam-filters  # create a new branch pointing to the last commit, so you don't lose it
git reset --hard HEAD~1  # move 'nonuniform' branch back one commit
git push -f origin nonuniform
```

rbiswas4:
@kbarbary @dannygoldstein 
@tholoien is trying to fit SALT model parameters to SNLS supernovae. Is the radial dependence branch (with an implementation for SNLS worth waiting for at the moment ie. in a day or two?) Otherwise, it will help us to go ahead with the original megacam 'average' filters for now, and use this later when it comes online! 

Also, are the average megacam filters not registered in SNCosmo because of this or am I simply missing this? Should we consider doing a PR for that if it is a disjoint issue? If so, could you point to some example code (for example used in registering the SDSS filters) that such a PR could be modeled on?

kbarbary:
@rbiswas4 radial-dependent bandpasses are on a backburner at the moment, although I am working on other aspects of sncosmo right now. I hope to finish the radial bandpass PR (among other things) by November.

For now, I'd just follow this guide: http://sncosmo.readthedocs.io/en/v1.3.x/registry.html#using-the-registry-to-achieve-custom-built-ins to use the average megacam filters locally.

Probably not worth a PR to register the average bandpasses because bandpass definitions should be moved out of the package before the next release.

rbiswas4:
Thanks @dannygoldstein . Also good to know, that registering average megacam filters as part of SNCosmo is likely to get outdated in a couple of months and so not worth it!

kbarbary:
Closing in favor of #173

PR #126, opened 2015-11-21
add LocalMagSystem & 'csp' magsystem built-in

kbarbary:
This supersedes #117 (commits squashed). Could still use some slight fixes, but this is the general idea. 

I've removed the CSP reader from this PR because it seems to dataset specific. It's still useful, but it might be better as part of kbarbary/sndatasets, e.g., the future `sncosmo.datasets` module.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -26,7 +26,7 @@ from . import io
 from .utils import download_file, download_dir
 from .models import Source, TimeSeriesSource, SALT2Source, MLCS2k2Source
 from .spectral import (Bandpass, read_bandpass, Spectrum, MagSystem,
-                       SpectralMagSystem, ABMagSystem, LocalMagSystem)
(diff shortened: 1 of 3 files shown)


dannygoldstein:
did you squash all the commits from #117 into 1 and then squash some new commits onto those to make this pr?

kbarbary:
Yes.

dannygoldstein:
probably should make the sndatasets repository part of the sncosmo organization

kbarbary:
I think we should just make it part of sncosmo. To make it really useful, we'll have to integrate pretty tightly with sncosmo conventions for bandpass names and zeropoints anyway.

kbarbary:
One remaining issue here is that `LocalMagSystem` isn't a great name. Perhaps `CompositeMagSystem` since it represents a system where the fundamental standard is (or can be) different for each bandpass?

Issue #99, opened 2015-07-29
better bandpass treatment


kbarbary:
The bandpasses included with sncosmo as "built-ins" are sometimes "down-sampled" and "truncated" if their sampling is too fine or they cover a large wavelength range. It would be good to be able to:
- store the raw bandpass definitions
- take care of down-sampling & truncation in software (when loading bandpass)
- add multiple versions of a bandpass while we're at it(?)

_edit:_ Having the raw (full) bandpass definitions will give better provenance as well.
### Background
- **Down-sampling** is reducing the wavelength resolution of the bandpass. This is done simply for performance reasons: when integrating a SN spectrum through a bandpass, we need to chose some wavelength grid on which to evaluate the spectrum and the bandpass. Currently, we just use the wavelengths at which the bandpass is defined. This makes evaluating the bandpass simple - no interpolation necessary! However, the time it takes to do an integral scales linearly with the number of bandpass samples, so we don't want to sample more finely than necessary. In practice, a resolution of 5 Angstroms in wavelength is usually good (and is incidentally what SALT2/snfit uses).
- **Truncation** is reducing the wavelength coverage of the bandpass at low and high wavelengths. Practically, many bandpass definitions include a huge range of wavelengths where the transmission is zero or very nearly so, as a measurement of "out-of-band" transmission. This is of interest for some studies. For us, this wide wavelength coverage usually just hurts performance (simply because there are more samples; see above) and causes problems with model wavelength coverage: the model is only defined over a finite wavelength range, and we require that the bandpass definition not extend past the model. A wider bandpass makes this more difficult (or limits the range of redshifts where we can integrate the model through the bandpass).

PaulKuin:
For UV filters you need to have the red tail, as even 1% red leak, once the UV flux drops steeply, will give a substantial contribution. To deal with this you probably need a quick and dirty method to estimate that effect, then choose your sampling based on the effective band (the flux times transmission as a function of wavelength).

rcthomas:
Kyle, can you point to where you do the integral in the code?  Don't get excited, I probably don't have anything bright to say about it...

kbarbary:
@rcthomas See [this line](https://github.com/sncosmo/sncosmo/blob/c4b2bf7ae76a4ad3bd821a7b4ee378d2520cde75/sncosmo/models.py#L134).

[...]

Issue #10, opened 2013-08-08
Position-dependent Bandpasses


kbarbary:
There is currently no way to represent a filter whose transmission function depends on focal plane position, other than having many `Bandpass` objects to represent a single physical filter, e.g. `desg_0_0`, `desg_1_0`, etc.

The most natural solution is to implement a subclass of `Bandpass` that stores multiple transmission functions depending on position. We'll have to think about the details of the interface.

dannygoldstein:
Volunteering to work on this.

rbiswas4:
@dannygoldstein  Awesome! This would be very useful and get us up to speed on being able to analyze real data beyond average transmissions. 

Q1. Do you have a plan on @rcthomas design options:
> I think the design question here is whether to put a layer that generates Bandpass objects for any given x, y or make x, y an argument of a call to Bandpass

I tend to agree with making the bandpass layer approach (which also seems to be the @kbarbary's opinion. Has @kbarbary's opinion on this evolved? ), mostly because model methods only depend on the quantities like position through the bandpass. 

A possible case to keep in mind is to be able to assign transmission functions discreteley to ccds rather than a smooth interpolation in radius. 

Q2. Do you have a dataset in mind which you want to build as a test?

kbarbary:
Still haven't thought of any designs that I *really* like, but this has got me thinking about the problem again.

I'd be interested to hear any ideas Danny has. It's a deep-reaching change since Bandpasses are used everywhere in sncosmo. There's the additional problem of trying to maintain backwards compatibility if possible.

dannygoldstein:
The interface of `Bandpass` is set; rewriting it to implement `__call__` or `trans()` would require other core interfaces to be rewritten (via changes that percolate to `zpbandflux`, `bandflux`). This could affect backwards compatibility. 

If one does:

```python
>>> sncosmo.get_bandpass('megacam-i')
```

I think one should get back a subclass of `Bandpass` that looks something like:

```python
class DependentBandpass(Bandpass):

def at(self, x=None, y=None, ...):
      # compute wave, trans given x, y, etc. 
      return Bandpass(wave, trans)

@property
def trans(self):
    # return some average or characteristic transmission

@property
def wave(self):
    #return the wavelength that corresponds to trans
```

That way you can do something like:

```python
>>> sncosmo.get_bandpass('megacam-i').at(x=5., y=6.)
```

[...]

PR #128, opened 2015-12-28
Bandpass improvements

kbarbary:
This PR does two things to address #99:
- Make bandpasses callable continuous functions. One can still access `band.wave` and `band.trans`
  to get the arrays defining the bandpass. However one can now also get the
  the transmission at arbitrary wavelengths, using `band(wave)`. This means that bandpasses can
  be defined with arbitrary spacing without having a big effect on the speed and accuracy of synthetic photometry.
  
  Along with this change, there is a new option when initializing a Bandpass: `trunc_level`, which truncates out-of-band transmission. This is useful for working with bandpass data that is defined over a very wide range (and where most of the transmission is very close to zero). This lays the groundwork for using bandpass data as provided, without having to preprocess it.
- In Model.bandflux(), the above Bandpass capability is now used to sample the bandpass at a regular interval (of ~5 angstroms, for now), regardless of native bandpass sampling.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -128,15 +128,8 @@ def load_bandpass_microns(pkg_data_name, name=None):
 def load_bandpass_bessell(pkg_data_name, name=None):
     """Bessell bandpasses have (1/energy) transmission units."""
     fname = get_pkg_data_filename(pkg_data_name)
-    band = read_bandpass(fname, wave_unit=u.AA, trans_unit=u.erg**-1,
-                         name=name)
-
-    # We happen to know that Bessell bandpasses in file are arbitrarily
-    # scaled to have a peak of 1 photon / erg. Rescale here to a peak of
-    # 1 (unitless transmission) to be more similar to other bandpasses.
-    band.trans /= np.max(band.trans)
-
-    return band
+    return read_bandpass(fname, wave_unit=u.AA, trans_unit=u.erg**-1,
+                         normalize=True, name=name)
 
 
 def tophat_bandpass(ctr, width, name=None):
--- a/sncosmo/models.py
+++ b/sncosmo/models.py
@@ -89,10 +89,42 @@ def get_source(name, version=None, copy=False):
         return cp(registry.retrieve(Source, name, version=version))
 
 
+def _bandflux_single(model, band, time_or_phase):
+    """Synthetic photometry of model through a single bandpass.
+
+    Parameters
+    ----------
+    model : Source or Model
+    band : Bandpass
+    time_or_phase : `~numpy.ndarray` (1-d)
+    """
+
+    # Check that bandpass wavelength range is fully contained in model
+    # wavelength range.
+    if (band.wave[0] < model.minwave() or band.wave[-1] > model.maxwave()):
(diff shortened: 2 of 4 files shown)


dannygoldstein:
Any thoughts on how this will interface with #10?

kbarbary:
Now that I think about it, this might clash with #10 / #125 a bit: A non-uniform bandpass could still have a `at()` method that returns a `Bandpass`. However, with this PR creating a `Bandpass` now involves creating a spline, which is awkward and takes a bit of time. I might rework #125 onto this before merging to make sure the design is right.

There are a couple things I'm not sure about with the API: It seems like it would be better to call `band.transmission(wave)` rather than `band(wave)` - seems clearer to me. Any opinions?

I'm also not sure that `trunc_level` in the Bandpass initialization is the best way to handle unwanted out-of-band transmission. (It is better than what we do now though, which is to truncate the bandpasses outside sncosmo.)

rbiswas4:
The out of band transmission fix will be very useful.
This may be a place to bring up a couple of other issues to see if they fit into this (and the framework for radial dependence of bandpasses). 
- The atmospheric transmission is airmass dependent in a wavelength dependent way. Clearly, one way of accounting for this during analysis is using zeropoints for individual bands, which will obviously be there. But, I can imagine trying to use a prediction for such bandpasses, particularly during simulation (where this could be used to predict a scatter in zero points) and potentially used as extra information during analysis. Assuming that $T_{atm}(\lambda)$ is known, is there a good way to put this kind of thing in the framework extensions being considered? 
- While most of the flux calculation capabilities are extremely convenient (handles sampling on to required grids from native grids for the calculation behind the scene), can we also think of (a maybe incovenient) superfast way minimizing the number of interpolation operations (my belief is that the bottleneck of flux calculation operations is in interpolation/matching grids and checks) that we need to do? This has not been an issue for me so far, but might be helpful to outline for the future.

kbarbary:
@rbiswas4

- Assuming you can specify the bandpass transmission as a function of airmass, e.g., `T(wave, airmass)`, it would look the same as a "RadialBandpass" - it's a function of a single parameter. If you want the transmission to be a function of airmass and position, that would be something different. This work won't get us there, but will move us in the right direction. It might be good to know how LSST will be specifying transmission as a use case. My understanding is that LSST will be estimating per-observation total transmission functions. Is that right?

- Optimizing interpolations is a whole other issue, and one I'd like to revisit (after this bandpass stuff is done). You have to do at least some interpolation because you have to get the bandpass and the model on the same wavelength grid. To some extent you can use caching to minimize this. In fitting though, you're usually fitting `t0`, which means you still need to interpolate the model in phase on every iteration.

rbiswas4:
- OK, I just threw this in to make sure that there is a similar path to the airmass issue. So this is good. About LSST  specifying the transmission function, is your question about data products that will be available after data is taken? I have a vague recollection that this is the case, but I will try to confirm the status of the current plans soon, after AAS.
- Sure, we cannot completely avoid interpolation. But figuring out a procedure that minimizes this need would be good. (Also phase interpolations might become necessary for z fits)

kbarbary:
Closing, as this has been superceded by other commits on master.

PR #130, opened 2016-01-11
Sfd98 mirror

### LINKED ISSUE #129: Link to SFD98 map broken
The link to the SFD98 MW dust map in the "Adding Milky Way dust" example (http://sncosmo.readthedocs.org/en/latest/models.html#adding-milky-way-dust) is broken. Is there any mirror for it?

I found it on this page: http://lambda.gsfc.nasa.gov/product/foreground/f_products.cfm
but I'm not sure whether the file is the right one, e.g. it uses HEALPix with nside=512 while - guessing from the file name in the documentation - the expected file might be nside=4096.

ufeindt:
Fixes #129 

Since linking to the directory gives a 404, I just added 4 links.
I hope the markup is OK, github doesn't show the note block.

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -228,13 +228,13 @@ First, load the dust map (do this only once)::
    This supposes that you've downloaded the full resolution E(B-V)
    maps from and placed them in the given directory ``"/path/to/dust/maps"``:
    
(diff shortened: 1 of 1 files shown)


kbarbary:
Looks good, thanks! The markup looks right.

FYI, I usually build docs locally first to make sure I get the markup right. Which sphinx installed, it should just be `make html` in the `docs` directory, then point your browser to the `docs/_build/html` directory.

PR #150, opened 2016-07-14
Make `mcmc_lc` accept log priors, not prior pdfs, to avoid underflow

dannygoldstein:
This PR alters the interface of the `mcmc_lc` method, making its `priors` argument take a dictionary of functions that return the log prior, rather than the prior pdf. Reasons for doing this:
- computing the log probability directly is often cheaper than computing the pdf and then taking the log 
- it avoids underflow (no situations where an exception will be raised by trying to take the log of 0)


kbarbary:
This looks like a good idea. But we should make it backwards-compatible. How about adding a separate keyword like `lnpriors` or `logpriors` that takes precedence over `priors` if specified?

PR #168, opened 2016-12-19
add data flux covariance

kbarbary:
This adds a representation for photometric data covariance and the ability to read salt2 (snfit) formatted "covmat" files. (It doesn't yet *use* covariance in fitting or sampling functions; subject of future PR.)

## Representation

I went back and forth on the best way to represent covariance. What I settled on is simple: that covariance is just another column in the astropy Table of observations, but instead of it being a single number for each observation, it must be an array of length N (where N is the number of observations). This makes it so that `data['cov']` is an array of shape `(N, N)`, which is nice.

The one issue that arises is with slicing. Suppose you have a data table `data` of length 20 and do `data[0:10]`. After slicing, `data['cov']` will have shape `(10, 20)` not `(10, 10)`... The `Table` class couldn't have known that the rows in the `'cov'` column refer to each other and that the slicing should also be applied to each row.

Currently, the impetus is on the user to manually slice the covariance matrix correctly. We could provide a "covariance-aware" slicing function and instruct users to use that rather than `data[...]`. For example:

```python
new_data = sncosmo.select(data, mask)  # instead of new_data = data[mask]
new_data = sncosmo.select(data, slice(0,10)) # instead of new_data = data[0:10]
```

Another alternative would be a custom user-facing class for photometric data that is a subclass of astropy.Table. But I wasn't sure we wanted to commit to that design yet.

## I/O

To read snfit "covmat" files, there is a new keyword `cov` in `read_lc()`:

```python
data = read_lc('lc-03D4ag.dat', format='salt2', cov=True)
```

This will expect a `COVMAT` keyword in the file and read the covariance matrix from that file, storing it in a column with name `'Cov'`. Perhaps the keyword and/or column name should be `covmat` rather than `cov` to match the snfit header keyword?


kbarbary:
Another question is whether the *default* should be to read the covmat file if `COVMAT` is present. That way, `read_lc(fname, format='salt2')` can be used on all data, regardless of whether it has covariance. This could potentially cause users' code to break though.

rbiswas4:
1. I think it would be good to standardize what is being read into the covariance matrix structure: Is it the covariance or relative correlation cov(i, j) / sigma(i) / sigma(j) so that when this gets used in fitting, the fitting code unambiguously knows what it is reading in. This choice can then be imposed by assertion statements in the code body (if it is relative, diagonal entries must be 1., if it is the full covariance, the diagonal entries should match the flux error squared. 
2. I agree that a default of not reading in the covariance would be best. (My personal choice would have be to keep this in a separate table altogether. Then this issue would be simple, and actually the slicing is simpler too.)

kbarbary:
@rbiswas4 I didn't consider even having "relative correlation" be an option. Is there any data provided that way? Currently this PR always expects the covariance to be exactly that: covariance on the flux values. Specifically, if covariance is present, the expectation is that the `fluxerr` column will be equal to the sqrt of the diagonal of the covariance, and that covariance should be used as-is in fitting.

If the default is to not read the covariance, and you want to analyze all the JLA data, you'll have to do some check to see if you expect the covariance to exist, like

```python
# decide if there will a covmat file based on filename:
expect_cov = re.match('lc-..D...\.dat', fname) is not None
data = read_lc(fname, format='salt2', cov=expect_cov)
```

Another option would be to make `cov=True` read the covariance *if present* and silently procede if not present. (Currently in this PR an error is raised if `cov=True` and covariance is not present.) This might be the best of both worlds: `cov` would still be False by default, but `cov=True` would Just Work<sup>TM</sup> on all salt2-format files.

> (My personal choice would have be to keep this in a separate table altogether. Then this issue would be simple, and actually the slicing is simpler too.)

I thought about this, but it seemed like a pretty big pain to have to pass around a covariance array alongside the data table and manually keep the two in sync. Also, `read_lc` would have to return a 2-tuple of `data, cov` when `cov=True` was specified. What would the benefit be? How would slicing be simpler?

rbiswas4:
@kbarbary 
> I didn't consider even having "relative correlation" be an option. Is there any data provided that way?

I am not a hundred percent sure if I have seen this for light curve data, but people in our field seem to use both covariance and relative covariance defined is `cov(i, j) / sigma(i) / sigma(j)` frequently. So, my guess is it will show up sooner or later. I think your expectations are perfectly reasonable (and if the data is in relative covariance format, it is fine to expect the user to format their data correctly) but should be mentioned explicitly, and perhaps even enforced by some checks. 

As an aside, I think it should be pretty easy to use the data covariance (since we already use a model covariance) we are already in the right format.


 > I thought about this, but it seemed like a pretty big pain to have to pass around a covariance array alongside the data table and manually keep the two in sync.  Also, read_lc would have to return a 2-tuple of data, cov when cov=True was specified.  What would the benefit be?

This is true. If this is an idea that we had to go with, we would have to promote this to a class with the covariance being an attribute. Carrying around two tables as a tuple will be madness as you point out.

I think the main benefits would be very explicit data structures. When I see a table, I want column-wise operations to be sensible. Also, then we know how we to optimally write these structures to disc, unlike writing the matrix as a NXN table we will be able to pack it.

 > How would slicing be simpler?
Let us say we choose the slice `s = [0, 3, 4, 7, 10]` for some reason. Then we can do

```
l = lc[['time', 'band', 'flux', 'fluxerr', 'zp', 'zpsys']][s], 
c = cov[s, s] 
```
to get the both the light curve and covariance.  (I might be using `pandas` syntax rather than `astropy` because I am more familiar, but similar things can be done).

Maybe similar things can be done easily enough in the method you propose?

kbarbary:
You actually would need to do:
```
c = cov[np.ix_(s, s)]
```
because `c = cov[s, s]` gives you a 1-d array of shape `(5,)` not a 2-d array of shape `(5, 5)`.

With the current PR, you'd have to do:
```python
d2 = data[['Date', 'Flux', 'Fluxerr', 'ZP', 'Filter', 'MagSys']][s]
d2['Cov'] = data['Cov'][np.ix_(s, s)]
```
or 
```python
d2 = data[s]
d2.replace_column('Cov', d2['Cov'][:, s])
```

though in astropy 1.3, I think this can be:
```python
d2 = data[s]
d2['Cov'] = d2['Cov'][:,s]
```

I agree that creating a custom subclass of astropy.Table would be nicer in some aspects; I'm just afraid of unforeseen pitfalls with that approach. My thought was that if *all* we want is to get the slicing right, a simple convenience function could do that.

We can always start returning a custom astropy.Table subclass in the future if we decide that it is indeed a good idea.

rbiswas4:
Sure. I agree that this gives us all of the functionality we need. We can iterate on subclasses and upgrade this stuff as required.

kbarbary:
OK, I changed the keyword to mean "read covariance matrix if available". Demo:

```python
>>> data = sncosmo.read_lc('lc-03D4ag.list', format='salt2', read_covmat=True)

>>> data.colnames
['Date', 'Flux', 'Fluxerr', 'ZP', 'Filter', 'MagSys', 'Fluxcov']

>>> data['Fluxcov'].shape
(62, 62)
```

Note that the keyword is `read_covmat` and the column name is `"Fluxcov"` (to match "Fluxerr"). Not sure if these are the best name choices though.

kbarbary:
Added a `sncosmo.select_data()` function that handles the covariance matrix.

PR #173, opened 2017-01-19
radially dependent bandpasses

### LINKED ISSUE #10: Position-dependent Bandpasses

kbarbary:
This adds support for the radially dependent bandpasses that snfit uses (for e.g., the MegaCam filters). A short demo:

```python
In [2]: band = sncosmo.get_bandpass('megacampsf::g', 10.0)  # get bandpass at radius of 10 cm

In [3]: band([4000., 4500., 5000., 5500., 6000.])
Out[3]: array([ 0.00118021,  0.48407073,  0.49770138,  0.36358558,  0.        ])
```

A note on vocabulary: `Bandpass` (and subclasses such as the new `AggregateBandpass` introduced in this PR) are *always* simple 1-d functions with no radial dependence. The thing that implements the radial dependence is called a `BandpassInterpolator` (not a great name, but will do for now). Given a position, it returns to you a `Bandpass` (or subclass thereof). So:

- `Bandpass` (and subclasses): thing used throughout sncosmo
- `BandpassInterpolator`: creates `Bandpass`es (and subclasses)

Closes #10.


dannygoldstein:
👍

PR #202, opened 2017-05-14
Fixing inverse

astrobarn:
Changing inverse of covariance matrix to pseudo inverse to increase numerical stability. Also robust when errors are zero.


kbarbary:
:+1:

PR #214, opened 2018-04-24
Adding Hounsell et al. model to registry.

rbiswas4:
Adding Hounsell et al. model to registry.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -594,6 +594,17 @@ _SOURCES.register_loader('salt2-extended', load_salt2model,
                          args=('models/snana/salt2_extended',), version='1.0',
                          meta=meta)
 
+# SALT2 H17
+meta = {'type': 'SN Ia',
+        'subclass': '`~sncosmo.SALT2Source`',
+        'url': 'http://sdssdp62.fnal.gov/sdsssn/SNANA-PUBLIC/',
+        'note': "extracted from SNANA's SNDATA_ROOT on 24 April 2018. SALT2"
+        " model with wide wavelength range, Hounsell et al. 2017"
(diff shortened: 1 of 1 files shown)


kbarbary:
Looks good. Can you fix the pycodestyle error? (line too long; see travis build log)

rbiswas4:
OK, fixed style error and moved the reference to its own key. Since the data directory is up, I checked that this is downloading, but got an error in unpacking the tar into an unexpected directory. This is not my usual sncosmo setup, so something may be amiss in the setup).

rbiswas4:
The appveyor version seems to fail on astropy numpy dependence (requiring a different version of numpy).

kbarbary:
Ah, so the code expects that `salt2_wfirst-h17.tar.gz` will contain a single directory `salt2_wfirst-h17`, and fails otherwise. What's actually in the tarball?

Yeah, the appveyor tests need to be updated. If you could do it as part of this PR, great. Otherwise, no worries. Could just remove [these lines](https://github.com/sncosmo/sncosmo/blob/master/appveyor.yml#L21-L26) and bump the last Python version to 3.6.

rbiswas4:
Right now: I have `salt2_wfirst-h17.tar.gz` unpack to `salt2-h17` (which is the model name). This is what I thought the code expects. However, it is easy to change this to `salt2_wfirst-h17`.

rbiswas4:
So, made changes to both this and the data repository according to your comments on both these repositories:
- `salt2-h17.tar.gz` should contain `salt2-h17` in the data repository
- `sncosmo` builtins now has loaders with model names `salt2-h17`, all mention of `wfirst` or different names are dropped
- appveyor.yml changed to remove lines not desired anymore.

PR #217, opened 2018-06-13
add ztf transmission functions

dannygoldstein:
Add in the ZTF transmission functions from @ufeindt


kbarbary:
Thanks! Can you actually add the data files to https://github.com/sncosmo/sncosmo.github.io/tree/master/data/bandpasses ?

kbarbary:
Are these available on a webpage somewhere @ufeindt ? Would be nice to have a URL reference if possible.

ufeindt:
Right now these particular files are not available and I don't know of any other files floating around. I will suggest that they be made available on the official ZTF website.

kbarbary:
Thanks Uli! We can put them in sncosmo now and add a URL reference later when there is one.

In case it wasn't clear, with the exception of the Bessel filters, bandpass data isn't distributed with the package... it's hosted on sncosmo.github.io and downloaded as needed. So that's why the bandpass files should be added there, not here.

dannygoldstein:
this should be merged, the tests that fail are not a result of this PR

PR #218, opened 2018-06-13
add wfirst filter transmissions

dannygoldstein:
Add in wfirst filter transmission functions from rebekah hounsell


kboone:
@dannygoldstein, are you still wanting to merge this PR? It looks like the filter transmissions should go in the sncosmo.github.io repo.

benjaminrose:
What needs to happen with this PR? I’m willing to take over adding the Roman filters of that is needed.

kbarbary:
@benjaminrose The filter files themselves should be added to https://github.com/sncosmo/sncosmo.github.io. Once that is merged, this can be updated to use the filter files from that repo, following the convention of most other filters.

benjaminrose:
Thanks.

The Roman filter definitions will still change. How does SNCosmo handle updates to transmission curves?

benjaminrose:
We should also update the name to `roman-wfi` since the telescope is now named and has two instruments. 

@dannygoldstein, are you able to update this PR if I add the filters to sncosmo.github.io, or is it better to create a new PR?

dannygoldstein:
Happy to update, ping me when ready.

benjaminrose:
Question for anyone interested. NASA and STScI name all the filters with an "F" prefix. So "Y106" would be "F106". Do we want to conform to what NASA and others do, or keep it in the more vernacular? Either way, we should definitely comment on this in the documentation.

kboone:
> Thanks. The Roman filter definitions will still change. How does SNCosmo handle updates to transmission curves?

The `sncosmo` registries support versions, so I believe that you could add new transmission curves with updated version numbers if they change.

> Question for anyone interested. NASA and STScI name all the filters with an "F" prefix. So "Y106" would be "F106". Do we want to conform to what NASA and others do, or keep it in the more vernacular? Either way, we should definitely comment on this in the documentation.

We should use whatever name NASA/STScI use. The `sncosmo` convention is to use all lowercase filters, so I think that "f106" is the right filter name for your example.

sncosmo/builtins.py
```diff
@@ -71,6 +71,12 @@ def load_bandpass_bessell(pkg_data_name, name=None):
                          normalize=True, name=name)
 
 
+def load_bandpass_wfirst(pkg_data_name, name=None):
+    fname = get_pkg_data_filename(pkg_data_name)
+    return read_bandpass(fname, wave_unit=u.AA,
```
benjaminrose, line 76: The data files I am creating are now in microns.

sncosmo/builtins.py
```diff
@@ -404,8 +410,24 @@ def tophat_bandpass_um(ctr, width, name=None):
...
+    'filterset': 'wfirst',
+    'retrieved': '13 Jun 2018',
+    'description': 'WFIRST filters from Hounsell et al. 2017 ' 
+    '(http://adsabs.harvard.edu/abs/2017arXiv170201747H)'}
+for name, fname in [('j129', 'wfirst/WFIRST_J129.dat'),
+                    ('w149', 'wfirst/WFIRST_W149.dat'),
+                    ('y106', 'wfirst/WFIRST_Y106.dat'),
+                    ('z087', 'wfirst/WFIRST_Z087.dat'),
+                    ('f184', 'wfirst/WFIRST_F184.dat'),
+                    ('h158', 'wfirst/WFIRST_H158.dat')]:
```
benjaminrose, line 424: The naming convention is updated, see the website PR, https://github.com/sncosmo/sncosmo.github.io/pull/23.

srodney:
Pinging here to re-raise this PR.  This is a feature we'd like to use for ongoing Roman SN SIT work.  Looks like @benjaminrose  has a nearly-merge-able PR, but the hold-up may be a trivial conflict in builtins.py, if I'm reading it right...

 @kboone : is there more needed to get this approved?

kbarbary:
@srodney I think we'd still need to update this PR to remove the filter files and make the changes @benjaminrose pointed out. It may be simpler to just open a new PR.

benjaminrose:
@dannygoldstein, would you like to rework this into a new PR? If not, I can do it. I will even use [multi-author commits](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) to ensure you get credit for your hard work.

dannygoldstein:
Go for it Ben, I'm out of the SN game

PR #222, opened 2018-10-22
SNEMO model

cmsaunders:
This adds the SNEMO model (Saunders et al. 2018) to the models in sncosmo. It also adds a function for loading 2-d multi-vector ascii data in the format "x0, x1, y0, y1, y2....".

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -723,11 +723,10 @@ for name, file, ver in [('snemo2', 'snemo2_ev.dat', '1.0'),
     meta = {'type': 'SN Ia', 'subclass': '`~sncosmo.SNEMOSource`',
             'url': 'https://snfactory.lbl.gov/snemo/',
             'reference': 'Saunders et al. 2018'}
-    version = int(name[5:])
 
     _SOURCES.register_loader(name, load_snemomodel,
                              args=['models/snemo/'+file],
-                             version=version, meta=meta)
+                             version=ver, meta=meta)
 
 # =============================================================================
 # MagSystems
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -25,7 +25,7 @@ from . import io
 from . import snfitio
 from .utils import download_file, download_dir, DataMirror
 from .models import (Source, TimeSeriesSource, SALT2Source, MLCS2k2Source,
-                     _SOURCES)
+                     SNEMOSource, _SOURCES)
 from .bandpasses import (Bandpass, read_bandpass, _BANDPASSES,
                          _BANDPASS_INTERPOLATORS)
 from .spectrum import Spectrum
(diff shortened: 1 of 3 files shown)


rbiswas4:
Great to see this!

sncosmo/builtins.py
```diff
@@ -710,6 +710,25 @@ def load_mlcs2k2(relpath, name=None, version=None):
...
 
+
+# SNEMO
+def load_snemomodel(relpath, name=None, version=None):
+    abspath = DATADIR.abspath(relpath)
+    return SNEMOSource(abspath, name=name, version=version)
+
+for name, file, ver in [('snemo2', 'snemo2_ev.dat', '1.0'),
+                        ('snemo7', 'snemo7_ev.dat', '1.0'),
+                        ('snemo15', 'snemo15_ev.dat', '1.0')]:
```
kbarbary: It seems like it makes sense to call these `'snemo'` version `'2'` (or `'2.0'`), `'7'` and `'15'` rather than version `'1.0'` of three different models.

EDIT: Ah I see now that they're sort of different models.

sncosmo/io.py
```diff
@@ -104,6 +104,41 @@ def read_griddata_ascii(name_or_obj):
...
+    x1 : numpy.ndarray
+        1-d array.
+    y : numpy.ndarray
+        3-d array of shape (len(y0, y1, ...), len(x0), len(x1)).
+    """
+
+    if isinstance(name_or_obj, six.string_types):
+        f = np.loadtxt(name_or_obj)
+    else:
+        f = name_or_obj
```
kbarbary: I think this `if... else` isn't necessary since `np.loadtxt` handles either a file name or a file object. (and indeed the code as is wouldn't work if `name_or_obj` is a file object.

I usually only use the variable name `f` for file objects, but here it is an array. So I'd rename `f` to `data` or something, for consistency with the surrounding code.

sncosmo/models.py
```diff
@@ -955,6 +956,70 @@ def _flux(self, phase, wave):
...
+
+    Parameters
+    ----------
+    fluxfile : str or obj, optional
+        Filename of an ascii file containing 2-d
+        array of spectral flux density values for a grid of phase
+        and wavelength values. Assuming columns ``phase``, ``wavelength``,
+        ``e_0``, ``e_1``, ``e_2``...
+    n_vector : int
+        One of 2, 7, 15 to select SNEMO2, SNEMO7, SNEMO15
```
kbarbary: Update the docstring? (`n_vector` not actually an argument to the class.)

sncosmo/models.py
```diff
@@ -955,6 +956,70 @@ def _flux(self, phase, wave):
...
+    """
+    
+    def __init__(self, fluxfile='SNEMO7_ev.dat', name=None,
+                 version=7):
+
+        n_vector = int(float(version))
```
kbarbary: It seems like it would be better to determine `n_vector` from the shape of the `values` array after parsing the `fluxfile`. This would ensure consistency between the file contents and `n_vector`. 

In sncosmo, usually the `version` argument to the model matches the `version` in the registry (`1.0` in this case) and doesn't actually affect model behavior - it's just for human reading so you know which model you loaded.

sncosmo/models.py
cmsaunders: Oops, removed this.

sncosmo/models.py
cmsaunders: Yes, that is more logical. I went back and forth on how to support the three models initially, but an explicit key is not necessary.

sncosmo/io.py
cmsaunders: Okay, I implemented this.

cmsaunders:
Thanks for the code review!

PR #224, opened 2019-01-06
Include cython source in the source distribution

mwcraig:
This would make building sncosmo on Python 3.7 possible in conda-forge (see https://github.com/conda-forge/sncosmo-feedstock/pull/6#issuecomment-451778463).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,7 +3,7 @@ include CITATION.md
 include LICENSE.rst
 include setup.cfg
 
-recursive-include sncosmo *.c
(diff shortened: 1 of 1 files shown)


kbarbary:
Hey Matt! Thanks for this and sorry for the delay. Question: This will cause cython to be a hard requirement for install via pip source, whereas before it wasn't required for source builds. Is that basically a non-issue now do you think? Or should I change `setup.py` to avoid this somehow?

The failing tests are probably unrelated and I'm working on fixing them in #223, after which I do a new release version.

PR #229, opened 2019-02-08
P18 Models 

jpierel14:
This adds the extrapolated CC SEDs and SALT2 model from Pierel et. al 2018, using existing functions as they were designed to work within the SNCosmo framework.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -736,58 +736,59 @@ for name, file, ver in [('snemo2', 'snemo2_ev.dat', '1.0'),
 
 # P18
 p18Models_CC = [('snana-2004fe', 'SN Ic', 'CSP-2004fe.SED'),
-          ('snana-2004gq', 'SN Ic', 'CSP-2004gq.SED'),
-          ('snana-sdss004012', 'SN Ic', 'SDSS-004012.SED'),  # no IAU name
-          ('snana-2006fo', 'SN Ic', 'SDSS-013195.SED'),  # PSNID
-          ('snana-sdss014475', 'SN Ic', 'SDSS-014475.SED'),  # no IAU name
-          ('snana-2006lc', 'SN Ic', 'SDSS-015475.SED'),
-          ('snana-2007ms', 'SN II-pec', 'SDSS-017548.SED'),  # type Ic in SNANA
-          ('snana-04d1la', 'SN Ic', 'SNLS-04D1la.SED'),
-          ('snana-04d4jv', 'SN Ic', 'SNLS-04D4jv.SED'),
-          ('snana-2004gv', 'SN Ib', 'CSP-2004gv.SED'),
-          ('snana-2006ep', 'SN Ib', 'CSP-2006ep.SED'),
-          ('snana-2007Y', 'SN Ib', 'CSP-2007Y.SED'),
-          ('snana-2004ib', 'SN Ib', 'SDSS-000020.SED'),
-          ('snana-2005hm', 'SN Ib', 'SDSS-002744.SED'),  # PSNID
-          ('snana-2006jo', 'SN Ib', 'SDSS-014492.SED'),  # PSNID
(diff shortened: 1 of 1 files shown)


sncosmo/builtins.py
```diff
@@ -733,6 +733,62 @@ def load_snemo(relpath, name=None, version=None):
                              args=['models/snemo/'+file],
                              version=ver, meta=meta)
 
+
+# P18
+p18Models_CC = [('snana-2004fe', 'SN Ic', 'CSP-2004fe.SED'),
```
kbarbary: These look like all the same names as the SNANA CC models (e.g. L628). How do they relate to the SNANA versions? Do you want to change the prefix of the model names here? (`snana-2004fe` -> `<prefix>-2004fe`)

sncosmo/builtins.py
```diff
@@ -733,6 +733,62 @@ def load_snemo(relpath, name=None, version=None):
...
+          ('snana-2007lx', 'SN IIP', 'SDSS-018734.SED'),
+          ('snana-2007og', 'SN IIP', 'SDSS-018793.SED'),
+          ('snana-2007ny', 'SN IIP', 'SDSS-018834.SED'),
+          ('snana-2007nv', 'SN IIP', 'SDSS-018892.SED'),
+          ('snana-2007pg', 'SN IIP', 'SDSS-020038.SED')]
+
+ref = ('SNSEDExtend', 'Pierel et al. 2018'
+       '<https://arxiv.org/abs/1808.02534>')
+for name, sntype, fn in p18Models_CC:
+    relpath = os.path.join('models','pierel','cc_models',fn)
```
kbarbary: Run the `checkstyle` script in the repo locally.

kbarbary:
Thanks for adding this! Generally looks good pending some changes in the data repository and the question about model names.

sncosmo/builtins.py
jpierel14: I can, up to you how you want to organize the models. These are all extensions of the existing SNANA models, with the ugriz bands untouched, the y band updated for the PLASTICC competition, and the JHK bands added based on the paper. I can keep them separate, but since the optical templates are really the same it feels a little like duplicating the repositories. Maybe we should just combine them somehow?

kbarbary: Ah, I see. Right now there are errors due to trying to register a model with the same name and version as an existing one, so we need to either change the name or the version. Sounds like it might make sense to call these version 2.0 of the existing models, as they should be preferred in all cases. Does that make sense?

The only issue I foresee would be if someone was doing `sncosmo.Model(source='snana-2004fe')`, when they update sncosmo to a new version, they'll suddenly get wider wavelength coverage in that model. That might change their results or potentially cause an error. (To get the original model back, one can do `sncosmo.Model(source=sncosmo.get_source('snana-2004fe', version='1.0'))`). This was actually intended in the original design: you get the latest version by default and if you want a specific version, you specify it. But now I'm wondering how disruptive it would be to existing code. Any thoughts?

jpierel14: I see the concern with existing code. In most cases it won't be a problem because users won't have been accessing the near-IR wavelengths with their code, since they didn't exist. But I can see there could be an issue if, for example, someone specifically did something like set a bound as wave[-1], expecting it to be in the z or y band, and suddenly it shifts out to the K band. 

I'd say either we just make these version 2.0 and make sure the documentation shows that these are changing the wavelength range and some of the y band data, or maybe follow what you did for SALT2 and give every one of these a name like "snana-2004fe-extended". A little more cumbersome but safer. Would you rather I do that?

For one thing though, the SALT2-extended model that I've uploaded should be indeed version 2.0 of the existing one, as it is an updated version of Steve's old one that is in there at the moment.

kbarbary: Let's go with `version='2.0'` for all of them. If there is some major problem, we can always change it.

jpierel14:
Should be good now based on this conversation, and made changes to the data repo.

PR #231, opened 2019-02-14
Update citation in readme to v1.4.0

benjaminrose:


benjaminrose:
Never mind. Maybe it should be the all version DOI, 10.5281/zenodo.592747, rather than a specific version. Let me know and I can update it.

kbarbary:
Yeah now that all-version DOIs exist, that would be better. I think I also told @rbiswas4 to cite the [ASCL entry](http://www.ascl.net/1611.017) at some point. Honestly not sure which should be preferred.

benjaminrose:
I think ASCL is nice, especially its integration with ADS, but it also just points to a DOI citation. There is always the option to write a "paper" for [JOSS](https://joss.theoj.org). It does make a new citation option, but one that is not version dependent.

I think my vote would be the all-version DOI.

kbarbary:
Sounds good, let's go with the all-version zenodo DOI as the "recommended" citation, but also link to the ASCL entry for reference.

benjaminrose:
I also just noticed the CITATION.md file. I'll update that file as well.

kbarbary:
Thank you!

Issue #240, opened 2019-06-04
HTTP Error 404: Not Found. requested URL: 'http://sncosmo.github.io/data/models/pierel/salt2.tar.gz'


tallamjr:
Recently we have been getting this error when running CI checks on [`snmachine`](https://github.com/LSSTDESC/snmachine):
```
test/snfeatures_test.py:79: in fit_templates
    nprocesses=nprocesses, seed=42)
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/snmachine/snfeatures.py:723: in extract_features
    self.model=sncosmo.Model(self.templates[mod_name])
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/models.py:1053: in __init__
    self._source = get_source(source, copy=True)
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/models.py:89: in get_source
    return cp(_SOURCES.retrieve(name, version=version))
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/_registry.py:179: in retrieve
    version=latest_version)
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/builtins.py:448: in load_salt2model
    abspath = DATADIR.abspath(relpath, isdir=True)
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/utils.py:466: in abspath
    download_dir(url, os.path.dirname(abspath))
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/utils.py:359: in download_dir
    _download_file(remote_url, buf)
../../../miniconda/envs/snmachine/lib/python3.6/site-packages/sncosmo/utils.py:236: in _download_file
    with closing(urlopen(req, timeout=timeout)) as remote:
../../../miniconda/envs/snmachine/lib/python3.6/urllib/request.py:223: in urlopen
    return opener.open(url, data, timeout)
../../../miniconda/envs/snmachine/lib/python3.6/urllib/request.py:532: in open
    response = meth(req, response)
../../../miniconda/envs/snmachine/lib/python3.6/urllib/request.py:642: in http_response
    'http', request, response, code, msg, hdrs)
../../../miniconda/envs/snmachine/lib/python3.6/urllib/request.py:570: in error
    return self._call_chain(*args)
../../../miniconda/envs/snmachine/lib/python3.6/urllib/request.py:504: in _call_chain
    result = func(*args)
[...]

PR #241, opened 2019-06-04
WIP : Adding partial solution to #240

rbiswas4:
This is towards a solution for the issue #240 reported by @tallamjr.  We changed the definition of `salt2-extended` version 2 in `builtins.py` to download the correct file. However, I still find 
```
 model = sncosmo.Model(source='salt2-extended')
Downloading http://sncosmo.github.io/data/models/pierel/salt2-extended.tar.gz
|========================================================================| 4.9M/4.9M (100.00%)         0s
  File "/Users/rbiswas/.local/lib/python3.7/site-packages/sncosmo-2.0.1-py3.7-macosx-10.7-x86_64.egg/sncosmo/models.py", line 1054, in __init__
    self._source = get_source(source, copy=True)
  File "/Users/rbiswas/.local/lib/python3.7/site-packages/sncosmo-2.0.1-py3.7-macosx-10.7-x86_64.egg/sncosmo/models.py", line 88, in get_source
    return cp(_SOURCES.retrieve(name, version=version))
  File "/Users/rbiswas/.local/lib/python3.7/site-packages/sncosmo-2.0.1-py3.7-macosx-10.7-x86_64.egg/sncosmo/_registry.py", line 177, in retrieve
    version=latest_version)
  File "/Users/rbiswas/.local/lib/python3.7/site-packages/sncosmo-2.0.1-py3.7-macosx-10.7-x86_64.egg/sncosmo/builtins.py", line 447, in load_salt2model
    abspath = DATADIR.abspath(relpath, isdir=True)
  File "/Users/rbiswas/.local/lib/python3.7/site-packages/sncosmo-2.0.1-py3.7-macosx-10.7-x86_64.egg/sncosmo/utils.py", line 464, in abspath
    raise RuntimeError("Tarfile not unpacked into expected "
RuntimeError: Tarfile not unpacked into expected subdirectory. Please file an issue.
```

	modified:   sncosmo/__init__.py
	modified:   sncosmo/builtins.py

### DIFF
--- a/sncosmo/__init__.py
+++ b/sncosmo/__init__.py
@@ -8,7 +8,7 @@ import os
 from astropy.config import ConfigItem, ConfigNamespace
 from astropy.config.configuration import update_default_config
 
-__version__ = "2.0.0"
+__version__ = "2.0.1"
 
 
 def test(package=None, test_path=None, args=None, plugins=None,
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -787,7 +787,7 @@ for name, sntype, fn in p18Models_CC:
 meta = {'type': 'SN Ia',
         'subclass': '`~sncosmo.SALT2Source`', 'ref': ref}
 _SOURCES.register_loader('salt2-extended', load_salt2model,
-                         args=('models/pierel/salt2',), version='2.0',
+                         args=('models/pierel/salt2-extended',), version='2.0',
                          meta=meta)
 # =============================================================================
 # MagSystems


kbarbary:
Yeah, it either needs to be re-tared to have a `salt2-extended` directory in the tarball, or we could rename the file on the server (sncosmo.github.io) to `salt2.tar.gz`. Either way, a PR to sncosmo.github.io is needed.
cc @jpierel14

rbiswas4:
Aha ... the tar needs to contain the correct directory name. Seems better to call it `salt2-extended`.

jpierel14:
> Yeah, it either needs to be re-tared to have a `salt2-extended` directory in the tarball, or we could rename the file on the server (sncosmo.github.io) to `salt2.tar.gz`. Either way, a PR to sncosmo.github.io is needed. The first option is probably better, just for clarity when looking at the data directory.
> cc @jpierel14

Pull request issued to change directory name

kbarbary:
This change should now work, now that the remote file has been fixed (thanks @jpierel14).

However, I noticed that we now have three "salt2-extended" type models:
- The original `salt2-extended` (version="1")
- `salt2-h17` (Hounsell et al 2017), added in #214 
- `salt2-extended` version="2" (Pierel 2018)

It's weird that they're not all `salt2-extended`, or it's weird that the third isn't `salt2-p18`. I'm not up-to-date on the models. Are the techniques of extending the templates close enough that they could be called three versions of the same model (and do later versions supercede earlier ones)? Or are they different enough that they should all have different names?

Since no one has actually used `salt2-extended` version="2" in a release (it was just released in v1.8 and has been broken up until now) we have a chance to correct this if we want to.

rbiswas4:
I think it would be good to make these consistent. All three are external extensions with the last two being prepared in the same way with different versions of the code. @jpierel14 can you help with the details? : https://github.com/jpierel14/SNSED_Repository

jpierel14:
@kbarbary I had to issue a second pull request, the download location of salt2-extended version 2 was still wrong somehow. Re the 3 different models:

@srodney Made version 1, which version 2 is a second iteration of so it's fine that version 1 is essentially replaced by version 2. Salt2-H17 is extremely close to version 2, and was made with the same methodology. However, salt2-H17 was used as part of Hounsell et. al 2018, while the methodology of salt2-extended version 2 was explicitly published in Pierel et. al 2018. They are extremely similar though, as per this plot:
Screen Shot 2019-06-05 at 2 18 19 PM

kbarbary:
Thanks for the explanation. Would it make sense to do:
- `salt2-h17` -> `salt2-extended` version 2
- `salt2-extended` version 2 -> `salt2-extended` version 3

or should we just leave it all as is?

My worry is that when you see `salt2-h17`, it's not really clear that it's a lot more like `salt2-extended` than it is like `salt2`. Or how it fits in relative to other `salt2-extended` models.

jpierel14:
I think it might make sense to leave them separate. That might mean we need salt2-extended-H17, to make it clear that it's closer to salt2-extended than salt2. However, making it version 2 and then P18 version 3 makes it look like they are 3 iterations of the same model. This is actually technically true, but slightly misleading.

Doing something like salt2-extended-H17 will mean that most people will default to salt2-extended version 2, which is probably the right choice, but can switch to H17 if they want to reproduce their work.

kbarbary:
Thanks for clarifying. Moving `salt2-h17` to `salt2-extended-h17` seems like a good improvement for understandability. I'll go ahead and merge this.

kbarbary:
Closing in favor of #242 as the version number didn't need to be incremented (2.0.0 hasn't actually been released)

Issue #246, opened 2019-07-08
Bandpass downloads from stsci not working


koconnor4:
Is it possible that the throughput website was updated recently? The urls in builtins.py for ACS & WFC3 bandpasses point to webpages without the data available causing it to download the webpage instead of the tables. 

It looks like the tables are located in zip files [here](http://www.stsci.edu/hst/instrumentation/wfc3/performance/throughputs), so maybe be best to just host these on sncosmo.github.io?

kbarbary:
Thanks for reporting! I've been traveling and won't be back with my laptop for a while. If someone is interested in giving this sooner, you could submit a PR to sncosmo/sncosmo.github.io adding the data files in the appropriate place under the `data/filters` directory and updating the `redirects.json` file to remove the redirects to the STScI site for the filters.

Alternatively, if anyone knows where they are moved to, we could just change the redirect address.

jpierel14:
WFC3 filters should be fixed by the pull requests above, ACS WFC filters still won't work, until someone can find the throughput tables.

kbarbary:
Hopefully fixed now. :crossed_fingers: Please check that it works for you!

PR #250, opened 2019-08-22
Fix for wfc3 filter throughput tables being moved on stsci website

jpierel14:
Fix for issue #246, see pull request on sncosmo.github.io for details.


jpierel14:
I see continuous integration problems here, not sure if it's my changes or the travis or something...let me know and I'll be happy to fix whatever is wrong.

sncosmo/builtins.py
```diff
@@ -90,7 +90,7 @@ def load_bandpass_remote_um(relpath, name=None):
 
 def load_bandpass_remote_wfc3(relpath, name=None):
     abspath = DATADIR.abspath(relpath)
-    _, wave, trans = np.loadtxt(abspath, unpack=True)
+    wave, trans = np.loadtxt(abspath, unpack=True)
```
kbarbary: Hmmm... did the format of the wfc3 files change? If so, the current released versions of sncosmo probably won't be able to load the new files now hosted at sncosmo.github.io.

jpierel14: Ah yes I forgot about that, the old files had an index column. Sorry about that I'll issue a new request, it'll have to be to both places unfortunately.

kbarbary: Already did the sncosmo.github.io one.

jpierel14: Oh you added back in the column? I just fixed load_bandpass_remote_wfc3 in this request, so if you fixed the files back in io then we should be good.

jpierel14: Other than it looks like you replaced the files, you didn't add a column right? So the few filters I added into wfc3-uvis aren't there anymore but are in builtins.py

jpierel14: I think the easiest thing to do here is just cancel this pull request. The new filters won't go in but that's okay, no other changes to builtins.py and sncosmo.github.io looks right.

kbarbary: Sorry for the confusion here.

I reverted the addition of all the files on sncosmo and instead added the same files that used to be hosted on the stsci website (I had them cached locally). These files have the three column format.

Going forward, to add the new filters and new versions of the old filters, we might want to do it in a separate sub-directory of sncosmo.github.io/data, so that old versions of sncosmo continue working. Do you think it is worth updating all the files? Or just adding the new ones? It's not clear to me what the difference is between the old and new versions, besides just formatting.

kbarbary:
I think this can be closed.

PR #251, opened 2019-09-25
Fix argument mutation in `fit_lc` and `mcmc_lc`

djperrefort:
I ran into a bug where the fitting functions would modify the `bounds` argument before copying it. For example:

```python
import sncosmo

model = sncosmo.Model('salt2')
data = sncosmo.load_example_data()
bounds = {'z':(0.3, 0.7)}
result, fitted_model = sncosmo.fit_lc(
    data, model,
    ['z', 't0', 'x0', 'x1', 'c'],
    bounds=bounds) 

print(bounds)
> {'z': (0.3, 0.7), 't0': (54995.0, 55180.0)}
```

This PR makes minor changes in `fit_lc` and `mcmc_lc` where instead of using

```python
if bounds is None:
    bounds={}
```

the bounds are copied as

```python
bounds = copy.deepcopy(bounds) if bounds else {}
```

It also adds a few tests to validate the above change.

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -439,8 +439,7 @@ def fit_lc(data, model, vparam_names, bounds=None, method='minuit',
     vparam_names = [s for s in model.param_names if s in vparam_names]
 
     # initialize bounds
-    if bounds is None:
-        bounds = {}
+    bounds = copy.deepcopy(bounds) if bounds else {}
 
     # Check that 'z' is bounded (if it is going to be fit).
     if 'z' in vparam_names:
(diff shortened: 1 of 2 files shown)


kbarbary:
Hey Daniel, sorry for the huuuuge delay looking at this. Started a new job in the Fall and it slipped off my radar. I'd like to get this into the next release.

Good catch! The main change looks good. The tests seem kinda overkill though and use unittest instead of pytest like the rest of the test suite. Can you instead just add a check to one of the existing fit tests, maybe around https://github.com/sncosmo/sncosmo/blob/master/sncosmo/tests/test_fitting.py#L62 ?

djperrefort:
My attention was focused elsewhere last week but it's back now. I'll simplify the test and switch to pytest sometime this week.

kbarbary:
No problem. I haven't yet found the time to do a release!

djperrefort:
Turns out the original fork for this PR was deleted. See PR #260 instead.

PR #253, opened 2019-11-26
Add SUGAR template in sncosmo 

FlMondon:
Thanks for your comment, you were right. You can find in this new pull request the updated version.


sncosmo/builtins.py
```diff
@@ -811,6 +811,38 @@ def load_snemo(relpath, name=None, version=None):
...
+        import requests
+        connect204 = requests.get('http://supernovae.in2p3.fr/sugar_template/sugar.tar.gz')
+        file_out_name = os.path.join(rep_name, 'sugar.tar.gz')
+        open(file_out_name, 'wb').write(connect204.content)
+        connect204.close()
+        os.system('tar -zxvf %s -C %s'%((file_out_name, rep_name)))
+        os.system('rm %s'%(file_out_name))
+
+    abspath = os.path.join(rep_name, 'sugar')
+    return SUGARSource(modeldir=abspath, name=name, version=version)
```
kbarbary: Instead of this implementation of `load_sugarmodel` you can do something very similar to `load_salt2model` in this file. (This requires adding a line to the `redirects.json` file in sncosmo.github.io, similar to [this one](https://github.com/sncosmo/sncosmo.github.io/blob/master/data/redirects.json#L10).)

This has several advantages:
- Takes care of untaring and unzipping
- Show download progress
- Generally more robust
- If the files ever move from the `supernova.in2p3.fr` server, we can update the `redirects.json` file and fix the problem immediately, instead of releasing a new version of sncosmo and requiring users to update.
- We don't use requests as a dependency
- Don't use `rm` or `tar` as these won't generally work on Windows.

sncosmo/models.py
```diff
@@ -573,6 +573,96 @@ def _flux(self, phase, wave):
...
+
+        mag_sugar += 48.59 # mag AB used in the training of SUGAR.
+        wave_factor = (wave ** 2 / 299792458. * 1.e-10)
+        return (self._parameters[0] * 10. ** (-0.4 * mag_sugar) /  wave_factor)
+
+    def bandflux_rcov(self, band, phase):
+        """
+        Model error is comming.
+        """
+        return np.zeros(phase.shape, dtype=np.float64)
```
kbarbary: I think it would be best to just remove this method until it is actually implemented.

kbarbary (dismissed):
Thanks for the modifications. There are a few minor changes to make. After sncosmo.github.io#18 is merged, you should be able to test this locally to make sure it is working as expected. It would also be nice to add some tests for the model, to ensure that it keeps working as expected in the future.

Could you or Pierre-François commit to supporting questions about the model that get raised from sncosmo users? It would be hard for me to do, as I'm not that familiar with the model.

FlMondon:
Thanks for your comment. Yes, I am working on the implementation of some test and for the questions about the model, we will support them no problem.

rubind:
Hey everyone, just checking to see if there are any updates. Thanks!

kbarbary:
@rubind Yes, just past the 1 year anniversary of the last comment!

@FlMondon I think this PR was close to complete but needed some tests. Is that right?

PFLeget:
Hi @kbarbary, @kboone, and @rubind  

I was looking if this pull request was done, which was not the case. I merged the recent updates from the `sncosmo` master branch and added tests for the `SUGARSource`. As @FlMondon added the comments from @kbarbary and that it looks like it passes all the requirement of `sncosmo`, is it ok now to push this branch into the master? If some stuff is still needed to merge this pull request, let me know :) 

Cheers, 
Pierre-François

sncosmo/models.py
```diff
@@ -573,6 +573,91 @@ def _flux(self, phase, wave):
...
+    of the model,``alpha_0``, ``alpha_1``, `alpha_2``, `alpha_3``, `CCM`` are the template vectors
+    of the model.
+
+    Parameters
+    ----------
+    modeldir : str, optional
+        Directory path containing model component files. Default is `None`,
+        which means that no directory is prepended to filenames when
+        determining their path.
+        m0file, alpha1file, alpha2file, alpha3file, CCMfile: str or fileobj, optional
```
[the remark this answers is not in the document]
PFLeget: I just fixed it with my last changes.

sncosmo/models.py
```diff
@@ -573,6 +573,91 @@ def _flux(self, phase, wave):
...
+                 alpha3file='sugar_template_3.dat',
+                 CCMfile='sugar_template_4.dat',
+                 name=None, version=None):
+
+        self.name = name
+        self.version = version
+        self._model = {}
+        self.M_keys = ['M0', 'ALPHA1', 'ALPHA2', 'ALPHA3', 'CCM']
+        self._parameters = np.zeros(len(self.M_keys))
+        self._parameters[0] = 1.0e10-15
```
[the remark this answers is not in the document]
PFLeget: It was fixed by Florian.

sncosmo/builtins.py
```diff
@@ -811,6 +811,23 @@ def load_snemo(relpath, name=None, version=None):
                              version=ver, meta=meta)
 
 
+#SUGAR models
+def load_sugarmodel(relpath, name=None, version=None):
+    abspath = DATADIR.abspath(relpath)
```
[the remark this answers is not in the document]
PFLeget: It was fixed by Florian.

sncosmo/models.py
PFLeget: It was removed by Florian.

sncosmo/builtins.py
PFLeget: It was done by Florian.

kboone:
These changes look good to me. I tried playing around with the SUGAR model and ran into one problem that needs to be addressed before we merge this. Outside of the valid range of the model, `model.flux()` gives you a large value for the flux instead of zero:

test_sugar_lc

It looks like this is because you are evaluating the model in magnitudes (and the model magnitude gets set to zero outside of the valid region) but then you transform it to a flux measurement and it becomes some non-zero value since 10**(-0.4 * 0) != 0. This is going to be a big problem for light curve fitting.

The easiest solution here is probably to explicitly set the flux to zero if you are outside of the valid phase range.

Also, is `Xgr` your preferred name for the amplitude parameter? It seems a bit confusing. `q0`, `A`, or `amplitude` would make more sense to me and be more consistent with other models, but I'll leave that decision to you.

PFLeget:
Hi @kboone, 

I implemented your suggestions. 

If you have any other suggestions before merging this pull request, let me know,
Cheers, 
Pierre-François

kboone:
Looks good to me and the tests are all passing. I'll merge this and put out a new release later today.

kboone:
The SUGAR model is now available in `sncosmo` v2.4.0. It should be out on conda and pypi shortly.

PFLeget:
Thank you so much for this :)

rubind:
Great work; thanks!

Issue #255, opened 2020-01-09
Package doesn't work with Astropy 4


timj:
As described in #225 sncosmo breaks with Astropy 4 because `astropy.extern` no longer includes `six`. I see that the code on master still tries to import `six`.

I couldn't find an explicit issue on this and I saw that #225 was closed.  I'm not entirely sure which approach you want to take to fix things but in theory I can do a pull request if you give me some guidance.

timj:
@rbiswas4  It looks like `six` was mostly removed but `astropy.extern.six` is still used in 3 test files.

kbarbary:
sncosmo v2.0 should have removed all `six` usage. A PR to clean up ones I missed would be appreciated! I can make a new release afterwards.

kboone:
I made a PR for this.

kbarbary:
solved by #258

PR #258, opened 2020-01-28
Remove all usage of the "six" package.

kboone:
This is a fix for #255. I removed the "six" package wherever it was used, and replaced the code with the Python 3 equivalent.


kbarbary:
Thanks Kyle!

PR #260, opened 2020-02-21
Fix argument mutation when fitting

djperrefort:
This PR makes minor changes in `fit_lc` and `mcmc_lc` where instead of using

```python
if bounds is None:
    bounds={}
```

the bounds are copied as

```python
bounds = copy.deepcopy(bounds) if bounds else {}
```

It also adds a few tests to validate the above change.


kbarbary:
Looks good except the tests are failing due to (I think) a missing import of `deepcopy`.

PR #261, opened 2020-02-24
PEP8 import optimizer

djperrefort:
I noticed a few unused import statements. This is potentially problematic because it clutters up the name space and adds overhead when importing `sncosmo`. The latter of these is especially important when running on systems like NERSC where imports are already painfully slow. I took the liberty of running an automatic PEP8 import optimizer.

NERSC is down for maintenance while they prepare for the Perlmutter machine, but I get a small ~1.6 % improvement in runtime when importing `sncosmo` on my local machine. The improvement is tiny, but it's essentially free.
Before optimization: 0.682 ms
After optimization: 0.671 ms

### DIFF
--- a/sncosmo/salt2utils.pyx
+++ b/sncosmo/salt2utils.pyx
@@ -13,16 +13,6 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from libc.math cimport fabs
 from libc.string cimport memcpy
 
-cimport
-numpy as np
-import numpy as np
-from cpython.mem import PyMem_Free, PyMem_Malloc
-from libc.math cimport
-
-fabs
-from libc.string cimport
-
-memcpy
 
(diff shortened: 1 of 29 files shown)


sncosmo/salt2utils.pyx
```diff
@@ -13,6 +13,16 @@ from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from libc.math cimport fabs
 from libc.string cimport memcpy
 
+cimport
+numpy as np
```
kboone: Looks like this broke all of the cimport statements

djperrefort: oops - Forgot to test against against a non-built install before pushing. I'll fix that quick.

djperrefort: Fixed it.

kbarbary:
Looks good!

PR #265, opened 2020-03-02
Spectra fitter

cmsaunders:


cmsaunders:
After discussing this with Kyle Barbary in January 2019, I waited too long to write this code, so I know that I forgot some of the things we decided in our conversation. There are a few things that I know need to be changed: `fit_sts` is an analog to `fit_lc`, but for spectral time series, but it's not a very nice name. `spectroscopic_data` is an analog to `photometric_data`, so I put it in the same file, but that doesn't make sense since the file is `photdata.py`. Some of the documentation for was copied and needs to be updated so that, for example, "band" is replaced by "wavelength." Please let me know everything else that I am forgetting.

kboone:
I think that we can close this PR now. Most of what was in this PR is included in #269.

Issue #266, opened 2020-03-02
Segfault in sncosmo.fit_lc


kboone:
I am running into a segfault with SALT2 fitting in sncosmo 2.1.0. When I call `sncosmo.fit_lc` on a specific light curve, I get a segfault. I have traced the offender down to the call to `m0 = self._model['M0'](phase, wave)` at line 692 of models.py which is calling `BicubicInterpolator.__call__` in `salt2utils.pyx`.

This bug is very strange. I'm fitting a lot of other light curves, and none of the others have this issue. It doesn't always crash at the same time... sometimes it gets 300 function evaluations in before crashing, other times it only gets 200 in. I assume that this must be some kind of memory corruption in `salt2utils.pyx`.

Evaluating model.bandflux() many times outside of the fit doesn't seem to cause crashes.

Here is some code that can be used to reproduce the crash on my machine. I get a segfault when running it in python, ipython or Jupyter notebook. I tried running the same code on a different machine with the same version of sncosmo and it doesn't crash. I'm not really sure how to debug this...

```
import sncosmo
from astropy.table import Table

sncosmo.register(sncosmo.Bandpass([4102., 5100.], [1., 1.], name='snfb'))
sncosmo.register(sncosmo.Bandpass([5200., 6289.], [1., 1.], name='snfv'))
sncosmo.register(sncosmo.Bandpass([6289., 7607.], [1., 1.], name='snfr'))

photometry = Table.read('./photometry.dat', format='ascii')
model = sncosmo.Model(source='salt2')

sncosmo.fit_lc(
    photometry,
    model,
    ['t0', 'x0', 'x1', 'c'],
    bounds={'x1': (-10, 10), 'c': (-1, 10), 't0': (-5, 5)},
    guess_t0=False,
    modelcov=True
)
```

and here is the photometry data (save it to photometry.dat)

```
band flux fluxerr time zp zpsys
snfb 4595752.543766973 382251.1484176252 -5.176698911830178 0.0 ab
snfv 5801808.934504561 288616.05353561015 -5.176698911830178 0.0 ab
snfr 5829701.398419394 206176.70295152883 -5.176698911830178 0.0 ab
snfb 6295441.3373373905 295957.034372337 -2.200020694998291 0.0 ab
snfv 10007962.019840367 234709.93911189566 -2.200020694998291 0.0 ab
snfr 10108232.969712026 156252.77488854234 -2.200020694998291 0.0 ab
[...]

PR #267, opened 2020-03-02
Don't segfault if a minuit fit fails.

kboone:
Fixes the segfault discussed in #266 

Note that this doesn't magically prevent fits from failing, it just prevents sncosmo from segfaulting, and minuit raises an exception instead.

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -37,6 +37,12 @@ def generate_chisq(data, model, signature='iminuit', modelcov=False):
     # parameters)
     if signature == 'iminuit':
         def chisq(*parameters):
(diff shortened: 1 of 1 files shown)


kbarbary:
Sounds like a reasonable fix, though ideally we could figure out why `BicubicInterpolator` segfaults with NaNs. Can you add a comment near the new code explaining why (could be short and just refer to `#266`)?

rbiswas4:
This is useful! Currently, it requires an extra try/except to dodge this, before checking for success.

kboone:
I added a comment explaining what is going on. I agree that `BicubicInterpolator` shouldn't segfault, so I think that we should leave #266 open for now. I'll look into it when I have time.

PR #269, opened 2020-05-20
Implement Spectrum class

kboone:
This PR implements a class to manipulate spectroscopic data.

This is currently a work-in-progress. See sncosmo/sncosmo-eps#3 for details.

### DIFF
--- a/sncosmo/specdata.py
+++ b/sncosmo/specdata.py
@@ -1,0 +1,169 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""Convenience functions for manipulating spectra."""
+
+import numpy as np
+from astropy.table import Table
+
+from .bandpasses import Bandpass
+from .photdata import PhotometricData
+
+__all__ = ['SpectrumData']
+
+
+def _estimate_bin_edges(wave):
+    """Estimate the edges of a set of wavelength bins given the bin centers.
+    This function is designed to work for standard linear binning along with
+    other more exotic forms of binning such as logarithmic bins. We do a second
+    order correction to try to get the bin widths as accurately as possible.
+    For linear binning there is only machine precision error with either a
+    first or second order estimate.
+    For higher order binnings (eg: log), the fractional error is of order (dA /
+    A)**2 for linear estimate and (dA / A)**4 for the second order estimate
+    that we do here.
+    Parameters
+    ----------
+    wave : array-like
+        Central wavelength values of each wavelength bin.
+    Returns
+    -------
+    bin_starts : `~numpy.ndarray`
+        The estimated start of each wavelength bin.
+    bin_ends : `~numpy.ndarray`
+        The estimated end of each wavelength bin.
+    """
+    wave = np.asarray(wave)
+
+    # First order estimate
+    o1 = (wave[:-1] + wave[1:]) / 2.
+
+    # Second order correction
+    o2 = 1.5*o1[1:-1] - (o1[2:] + o1[:-2]) / 4.
+
+    # Estimate front and back edges
+    f2 = 2 * wave[1] - o2[0]
+    f1 = 2 * wave[0] - f2
+    b2 = 2 * wave[-2] - o2[-1]
+    b1 = 2 * wave[-1] - b2
+
+    # Stack everything together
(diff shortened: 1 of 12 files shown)


sncosmo/specdata.py
```diff
@@ -0,0 +1,347 @@
...
+        """Return a list of bandpass objects for each wavelength element."""
+
+        dwave = 0.00001
+        bands = []
+
+        for bin_start, bin_end in zip(self.bin_starts, self.bin_ends):
+            bands.append(Bandpass(
+                [bin_start, bin_start + dwave, bin_end, bin_end + dwave],
+                [0., 1., 1., 0.],
+            ))
```
kbarbary, line 186: I *think* `Bandpass` has zero transmission outside the wavelength bounds, so I think you can simplify to
```python
Bandpass([bin_start, bin_end], [1.0, 1.0])
```

sncosmo/specdata.py
```diff
@@ -0,0 +1,347 @@
...
+
+    """
+    def __init__(self, wave=None, flux=None, fluxerr=None, fluxcov=None,
+                 bin_edges=None, bin_starts=None, bin_ends=None, time=None):
+        # Extract the bin starts and ends
+        bin_starts, bin_ends = _parse_wavelength_information(
+            wave, bin_edges, bin_starts, bin_ends
+        )
+        self.bin_starts = bin_starts
+        self.bin_ends = bin_ends
```
kbarbary, line 126: I'm curious about the benefit of allowing non-contiguous elements.

(If it were disallowed, I can see that `merge` would have to include a `rebin` operation).

kboone: I think that handling non-contiguous spectra is outside of the scope of `sncosmo`, so I'm going to remove this. Things like fitting are a lot more complicated to optimize if you allow non-contiguous spectra.

kboone:
I removed the rebinning and merging functionality. I think that they should probably be done with something like `specutils` instead and are outside of the scope of `sncosmo`.

kboone:
In the latest commit, I have implemented a new algorithm that properly handles sampling for spectral elements. I use the existing `integration_grid` function to figure out how to sample each wavelength element. This produces a list of wavelengths that can be used for computing bandpasses or models and a "sampling matrix" that the resulting array can be multiplied by to map it onto the spectral elements. I have an initial implementation for `bandflux` that works well. The sampling matrix can be stored as a sparse matrix, and it only needs to be computed once. This will be a lot faster than the previous implementation of using individual bands for each spectral element.

kboone:
The latest round of commits updates `fit_lc` to work with spectra.

kboone:
I have made several updates to `fit_lc` to support the majority of options to it (there are a lot!) and clean things up.

For guessing t0 and amplitude, I realized that there is no guarantee that spectra are sampled in the same way which complicates things. I implemented a method that generates synthetic photometry in tophat filters with 500A spacing for all spectra over the range of the model. The generated photometry can then be handled with the previous `guess_t0_and_amplitude` function. This is working well in my tests.

kboone:
@kbarbary I added tests and documentation for the Spectrum class. I'm not sure how to rebuild the docs so I haven't tested that out yet. Otherwise, this is pretty much good to go with everything discussed in the EP implemented.

kbarbary:
@kboone To build the docs locally you should be able to `make html` in the `docs` directory (assuming you have all the docs dependencies installed) and then look at them in `docs/_build/html`. Docs dependencies are listed [here](https://github.com/sncosmo/sncosmo/blob/master/.continuous-integration/travis/setup_dependencies_common.sh#L29-L33). This could be better documented!

kboone:
Thanks, I got that working and I updated the documentation to render nicely. That should be good to go now too.

kboone:
I just realized that the checks failed for the previous version of this PR because of some formatting issues. I cleaned that up. Is there anything else that you would like to see before it is merged @kbarbary? This implements everything that was discussed in the EP.

kboone:
Great, thanks! Do we need to do anything special to rebuild the docs?

kbarbary:
Nope, they build automatically on merges to master under the "latest" version. Looks like it built OK: https://sncosmo.readthedocs.io/en/latest/spectrum.html

PR #272, opened 2020-06-03
Added V19 templates

maria-vincenzi:
Added the list of 67 (x2) templates.
For each Core collapse there are two templates: one has been corrected for host dust extinction (e.g. V19_iPTF13bvn_HostExtCorr.SED), the other not (e.g. V19_iPTF13bvn_noHostExtCorr.SED).

We can either build the templates as I did and have two separate model with names: 'V19_iPTF13bvn_HostExtCorr' and 'V19_iPTF13bvn_noHostExtCorr'

Or build a single model 'V19_iPTF13bvn' and have two versions: version= '1.0_HostExtCorr' and version='1.0_noHostExtCorr'.

I'm not sure which is the nicer/clearer way of doing this.


kboone:
This looks great, and it would be very nice to have your models in `sncosmo`!

By default, `sncosmo` will load whichever template is the "latest version" given by an alphabetical comparison of the strings. Do you think that one set of models makes more sense to have as the default? Is it important to have both sets? Most users aren't going to look into the details...

To fit with the naming scheme of other models in `sncosmo`, I think that it would be good to rename the models to something like `v19-iptf13bvn`. See https://sncosmo.readthedocs.io/en/v2.1.x/source-list.html for the formatting of names of current models.

maria-vincenzi:
Thanks Kyle!
I think the default templates should be the one that are NOT corrected for dust extinction (i.e. the "original" core collapse event, as observed). All other CC templates currently in sncosmo are not corrected for host extinction. Therefore, setting the V19 default as the not-corrected templates as well would probably be the most coherent approach. I would prefer the user to "intentionally" select the templates that are corrected for dust extinction and than consider simulating additional dust using all the functions available in sncosmo.extinction. 

I implemented two versions of the templates:
version 1.1_HostExtCorr and version 1.2_noHostExtCorr (just to keep it more explicit)
I think we should update the documentation explaining all this!

maria-vincenzi:
I added a note and a couple of more details so that the documentation is updated and clear (hopefully..if not feel free to correct).

rbiswas4:
- Tried this out and the models work once I download the data to the appropriate directory. 
- To get this working with downloading will require a [merge to master](https://github.com/sncosmo/sncosmo.github.io/pull/22)

Thanks @maria-vincenzi ! It would be great to have these models as part of sncosmo.

@maria-vincenzi I wanted to bring up the question that @kboone asked: I would imagine that if I was either simulating or using a core collapse SN to match data, I would use the extinction corrected model (+ extinction appropriate for my host galaxy) . I seems that the only time one would want the uncorrected events would be if one wanted to use the real data of those SN, or maybe do an upgrade to your work with new SN. So following the default logic above, I would guess that the extinction corrected ones should be the default. But maybe I am missing something that I should be concerned about. Thoughts ?


A couple of comments on the usage of models in general for everyone:
- using the sources will require the use of SN names. While it is possible to see the type of a particular SN model, is there an easy way to pick out say SN Type II ?  (e.g for the nugent templates, we could see this from the model name:
```
s = sncosmo.get_source('nugent-sn2n')
```
If I want to jus get the models of a particular type e.g.  `SN II(P/L)`  is there a convenient way to do this ? Does it make sense to have a method to enable this (or to have an easily accessible label for any source) in `sncosmo`

kbarbary:
> All other CC templates currently in sncosmo are not corrected for host extinction

I believe the (now ancient) Nugent templates are corrected. Are really none of the others corrected? Ideally the models in sncosmo are corrected, but maybe it is time we be a bit more explicit about whether or not they are corrected, perhaps via a metadata field. Outside the scope of this PR though.

For the purposes of this PR, I think the corrected and un-corrected versions here should probably have different names (perhaps via a suffix like `-corr` or whatever). Increments in version are usually used to mean "the old version is outdated, use the latest" whereas here you might want to use the corrected or uncorrected version depending on your purpose: one doesn't replace the other.

> If I want to jus get the models of a particular type e.g. SN II(P/L) is there a convenient way to do this ? Does it make sense to have a method to enable this (or to have an easily accessible label for any source) in sncosmo

There isn't yet but it should be relatively easy to do this, again with metadata. Outside the scope of this PR though. Can you open a separate issue for that?

rbiswas4:
> For the purposes of this PR, I think the corrected and un-corrected versions here should probably have different names (perhaps via a suffix like -corr or whatever). 

I agree that this would be sufficient to merge the PR ! I would love to hear a confirmation (and have in documentations a recommendation of when to use the corrected and un-corrected versions) from @maria-vincenzi.


> here isn't yet but it should be relatively easy to do this, again with metadata. Outside the scope of this PR though. Can you open a separate issue for that?
Yep ... separate issue for discussion. And yes, I was able to use other `sncosmo` methods  to do this which would be good to push back as a method to `sncosmo`. Will file ...

maria-vincenzi:
> I believe the (now ancient) Nugent templates are corrected. Are really none of the others corrected? Ideally the models in sncosmo are corrected, but maybe it is time we be a bit more explicit about whether or not they are corrected, perhaps via a metadata field. Outside the scope of this PR though.

Yes you are probably right! Nugent's templates are "corrected". The SNANA templates and the SDSS templates (Sako 2011) are built from single/well-observed core collapse events and has not been corrected for dust extinction.
I'm happy to make the changes that Kyle Barbary suggested and assign version 1.1 to all templates and have different names for corrected and uncorrected templates  ('v19_2013ge', 'v19_2013ge_corr'). 

>  I would love to hear a confirmation (and have in documentations a recommendation of when to use the corrected and un-corrected versions)

In Vincenzi et al 2019 I explain this a bit more in details. The idea is that you have two options:
1) You can use the dust-corrected templates and apply dust extinction _separately_ choosing your favourite dust model(s) and your favourite dust extinction distribution. This allows you to "augment" the SED models and simulate a more diverse set of lightcurves.

2) You can use dust-uncorrected templates i.e. SED models that has not been corrected for reddening and still include the effects of the original host dust extinction. In this case you don't have to worry about applying dust extinction again.
This is the option you can use if you don't particularly trust the dust extinction corrections I applied and/or if you want to re-simulate the core collapse event _as originally observed_.

Hope this is clear. I can add this to the documentation (maybe using footnotes at the end of the table in the [List of Built-in sources](https://sncosmo.readthedocs.io/en/v2.1.x/source-list.html#)  page)

rbiswas4:
@maria-vincenzi Thanks! Yes, this is now clear.

sncosmo/builtins.py
```diff
@@ -750,6 +750,152 @@ def load_2011fe(relpath, name=None, version=None):
...
+                ('v19_2016gkg_corr', '1.0','V19_HostExtCorr', 'SN IIb', 'V19_SN2016gkg_HostExtCorr.SED'),
+                ('v19_2016gkg', '1.0', 'V19_noHostExtCorr', 'SN IIb', 'V19_SN2016gkg_noHostExtCorr.SED'),
+                ('v19_iptf13bvn_corr', '1.0','V19_HostExtCorr', 'SN Ib', 'V19_iPTF13bvn_HostExtCorr.SED'),
+                ('v19_iptf13bvn', '1.0', 'V19_noHostExtCorr', 'SN Ib', 'V19_iPTF13bvn_noHostExtCorr.SED')]
+
+note= """Templates from Vincenzi et al. 19.
+Each template is extended in the ultra-violet (1600AA) and in the near infrared (10000AA).
+Each template can be used in its original version (v19_sn-name) or in its host dust extinction corrected version (v19_sn-name_corr)."""
+for name, vrs, sub_ref, sntype, fn in V19_CC_models:
+    ref = (sub_ref, 'Vincenzi et al. 2019'
```
kbarbary, line 891: Why two different `sub_ref`s if both refer to the same paper?

sncosmo/builtins.py
```diff
@@ -750,6 +750,152 @@ def load_2011fe(relpath, name=None, version=None):
     _SOURCES.register_loader(name, load_timeseries_ascii,
                              args=(relpath,), version='2.0', meta=meta)
 
+# V19
+V19_CC_models = [('v19_asassn14jb_corr', '1.0','V19_HostExtCorr', 'SN II', 'V19_ASASSN14jb_HostExtCorr.SED'),
+                ('v19_asassn14jb', '1.0', 'V19_noHostExtCorr', 'SN II', 'V19_ASASSN14jb_noHostExtCorr.SED'),
```
kbarbary, line 755: nitpick: formatting. Indent like
```python
V19_CC_models = [
    ('v19_asassn14jb_corr', '1.0','V19_HostExtCorr', 'SN II', 'V19_ASASSN14jb_HostExtCorr.SED'),
    ('v19_asassn14jb', '1.0', 'V19_noHostExtCorr', 'SN II', 'V19_ASASSN14jb_noHostExtCorr.SED'),
    ...
]
```

sncosmo/builtins.py
```diff
@@ -750,6 +750,152 @@ def load_2011fe(relpath, name=None, version=None):
...
+                ('v19_asassn15oz_corr', '1.0','V19_HostExtCorr', 'SN II', 'V19_ASASSN15oz_HostExtCorr.SED'),
```
kbarbary, line 756: All existing sources use hyphens in their names rather than underscores, so it would best to keep that consistency. So `v19-asassn15oz-corr`.

kbarbary:
Bump. Just some minor changes required. We may do a version release soon, so it would be great to merge this before that happens!

rbiswas4:
@maria-vincenzi  What kind of interpolation have you been using with these models?

sncosmo/builtins.py
```diff
@@ -750,6 +750,155 @@ def load_2011fe(relpath, name=None, version=None):
...
+    ('v19-2016bkv', '1.0', 'SN II', 'V19_SN2016bkv_noHostExtCorr.SED'),
+    ('v19-2016gkg-corr', '1.0', 'SN IIb', 'V19_SN2016gkg_HostExtCorr.SED'),
+    ('v19-2016gkg', '1.0', 'SN IIb', 'V19_SN2016gkg_noHostExtCorr.SED'),
+    ('v19-iptf13bvn-corr', '1.0', 'SN Ib', 'V19_iPTF13bvn_HostExtCorr.SED'),
+    ('v19-iptf13bvn', '1.0', 'SN Ib', 'V19_iPTF13bvn_noHostExtCorr.SED')
+]
+
+note= """Templates from Vincenzi et al. 19.
+Each template is extended in the ultra-violet (1600AA) and in the near infrared (10000AA).
+Each template can be used in its original version (v19_sn-name) or in its host dust extinction corrected version (v19_sn-name_corr)."""
```
kbarbary, line 890: minor nit
```suggestion
Each template is extended in the ultraviolet (1600AA) and in the near infrared (10000AA).
Each template can be used in its original version (v19-sn-name) or in its host dust extinction corrected version (v19-sn-name-corr)."""
```

kbarbary (approved):

kbarbary:
@maria-vincenzi ready to merge?  The failing tests are unrelated.

maria-vincenzi:
Hi!
Yes! I was worried about all the failings but if you think that they can be ignored, we are ready to go!

maria-vincenzi:
> @maria-vincenzi What kind of interpolation have you been using with these models?
@rbiswas4 I used mainly gaussian processes (1d for LC fitting and 2d for the extrapolation of the full Flux(time, wavelength) surface). You can find all the details here: https://arxiv.org/abs/1908.05228

Also, I see the pull request is not merged yet. Is there anything else you want me to do/change? 
Cheers,
Maria

rbiswas4 (approved):

rbiswas4:
@maria-vincenzi  I think there is a plan by @kbarbary to merge a few PRs together and release. 
On the question about interpolation : What I had in mind was about simulations or fits that your model here would enable. In the current code setting, if we use your spectra as time series, it will get interpolated using cubic interpolation. It is easy to do calculations where we replace this by a linear interpolation (a GP will take some more work I think, but I agree that would be an improvement.)

I was wondering if you had a recommendation between using a linear or cubic interpolation from a scientific point of view. Thanks!

PR #286, opened 2020-12-03
Iminuit deprecation of "forced_parameters"

kboone:
In newer versions of iminuit, the "forced_parameters" keyword has been deprecated and replaced by "name". Right now, running sncosmo with a new version of iminuit spits out deprecation warnings for every fit.

It looks like this change happened in iminuit v1.4.3 (from June 24, 2020). We can approach this either by dropping compatibility with v1.4.2 or earlier, or by including a workaround in the code to use the right keyword depending on the iminuit version. This PR implements the workaround. There seem to have been quite a few changes to the iminuit API recently... I'm not sure if we are affected by any other ones.

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -668,8 +668,8 @@ def fit_lc(data=None, model=None, vparam_names=[], bounds=None, spectra=None,
         ndof -= len(vparam_names)
 
         # Minuit keywords changed in v1.4.3. Handle that gracefully.
-        from packaging import version
-        if version.parse(iminuit.__version__) >= version.parse("1.4.3"):
+        from distutils.version import LooseVersion
+        if LooseVersion(iminuit.__version__) >= LooseVersion("1.4.3"):
             param_names_kwargs = {'name': model.param_names}
         else:
             param_names_kwargs = {'forced_parameters': model.param_names}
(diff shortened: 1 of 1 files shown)


kboone:
There are some similar problems with emcee where we are currently using a deprecated API. In general, do we want to include workarounds to support old versions of these packages, or should we just have sncosmo work with the latest version?

kbarbary:
Good question. I think either way is OK. I guess if it is "easy" to support the deprecated API, then maybe it makes sense to just do it in case someone has a use case that requires an older version of the dependency. However, for most users, I would guess that if they are installing the very latest release version of sncosmo, then they won't mind also upgrading their dependencies to the latest version or a recent version.

For "strict" dependencies (in setup.py `install_requires`), with recent improvements to pip's dependency resolver, I feel better about just adding a lower bound to the dependency version in `install_requires` and it all working. For optional dependencies like iminuit and emcee, if we just switch to the new API we'd probably want to do a runtime version check and raise an exception if the installed version of the dependency is too old.

How old is the new emcee API?

kbarbary:
Looks like Python 3.5 doesn't have the `packaging` module built in. We could consider dropping support for Python 3.5 or some other work around, like parsing the version into a tuple of ints manually.

kboone:
Good catch. Looks like distutils has a similar function for version comparisons that has been around forever. I switched it to use that, which should let us maintain backwards compatibility with old versions of python 3.

kboone:
The entire `iminuit` API changed in v2.0 which breaks a lot of things and might make this PR redundant. See #291

kboone:
This PR was made redundant by #292 which overhauls the whole `iminuit` interface.

PR #287, opened 2020-12-03
Fixes to tox tests.

kboone:
This PR fixes a few issues with the current tests.

Currently tox doesn't run a bunch of tests because optional dependencies aren't installed (iminuit, emcee, nestle and matplotlib). I added those dependencies.

I also added a marker for "might_download". This marker is used in a lot of the tests, but wasn't previously labeled which spit out warnings.

Finally, I changed tox to run on whatever version of python 3 is available instead of forcing it to find python3.6 which I assume that most people no longer have on their machines.


kbarbary:
Looks good to me

PR #296, opened 2021-03-25
new SALT3 model

djones1040:


benjaminrose:
The CI failures are for code style and tests. Should be easily fixed.

djones1040:
thanks Ben!  Working on this now

djones1040:
this PR goes with https://github.com/sncosmo/sncosmo.github.io/pull/24 for the model files

sncosmo/models.py
```diff
@@ -997,6 +998,317 @@ def colorlaw(self, wave=None):
...
+                         'cdfile': cdfile, 'clfile': clfile}
+
+        # Make filenames into full paths.
+        if modeldir is not None:
+            for k in names_or_objs:
+                v = names_or_objs[k]
+                if (v is not None and isinstance(v, str)):
+                    names_or_objs[k] = os.path.join(modeldir, v)
+
+        # model components are interpolated to 2nd order
```
kbarbary, line 1088: I think of "2nd order" as quadratic, but this is using cubic interpolation. Is there a a wording mismatch?

sncosmo/models.py
```diff
@@ -997,6 +998,317 @@ def colorlaw(self, wave=None):
...
+            values *= self._SCALE_FACTOR
+            self._model[key] = BicubicInterpolator(phase, wave, values)
+
+            # The "native" phases and wavelengths of the model are those
+            # of the first model component.
+            if key == 'M0':
+                self._phase = phase
+                self._wave = wave
+
+        # model covariance is interpolated to 1st order
```
kbarbary, line 1100: "1st order" but using cubic interpolation?

sncosmo/models.py
```diff
@@ -997,6 +998,317 @@ def colorlaw(self, wave=None):
...
+                raise RuntimeError("Unexpected keyword: {}".format(words[i]))
+
+        # Set extinction function to use.
+        if version == 0:
+            raise Exception("Salt2ExtinctionLaw.version 0 not supported.")
+        elif version == 1:
+            self._colorlaw = SALT2ColorLaw(colorlaw_range, colorlaw_coeffs)
+        else:
+            raise Exception('unrecognized Salt2ExtinctionLaw.version: ' +
+                            version)
```
kbarbary, line 1286: Minor comment: `RuntimeError` might be the most appropriate exception here. Also, `version = 0` might indicate that the version keyword wasn't found in the file.

kbarbary (commented):
Cool. Can you add the class to the API reference docs in `docs/reference.rst`?

It looks like there's a fair amount of code duplication with `SALT2Model`, though I haven't taken a close look at the differences. Any thoughts on how we could reduce code duplication here?

sncosmo/models.py
djones1040: this is from the original SALT2 - I don't understand it but I'm hesitant to change it, given that it could indicate something about how the model is generated.  I removed this from the new code because I realized the only thing I really need to change is how the errors are generated, so I just made SALT3Source an instance of SALT2Source and overwrote the error method.

sncosmo/models.py
djones1040: see above

sncosmo/models.py
djones1040: I removed this code from SALT3Source (see comment above), but changed this in the SALT2Source just for fun

djones1040:
thanks so much for doing this review Kyle, just to summarize what I said above -  I realized the only thing I really need to change for SALT3 in sncosmo is how the errors are generated, so I just made SALT3Source an instance of SALT2Source and overwrote the error method in SALT3Source.  Seems to work well.

djones1040:
hey @kbarbary , just sending a reminder - think proposed changes have been made.  Thanks!

kboone:
The latest version looks good to me. If there are no objections, I'll merge it on Monday.

djones1040:
thanks so much @kboone - note PR https://github.com/sncosmo/sncosmo.github.io/pull/24 is also necessary for the data files to be included in the repo

sncosmo/models.py
kbarbary: That was probably me who wrote `Exception` initially! :smile:

kbarbary (approved):

kbarbary:
Looks good, sorry for the delayed re-review!

djones1040:
well crap, merging the new data repository allowed me to catch my error - SALT3 doesn't require an error scaling file, so it's best to not include a "dummy" file as I'd done in the past.  I edited the code to fix the code breaks that happened when the salt3 data were downloaded from the online repository and an "errscalefile" was not found.  Assuming this passes checks, my hope is everything will be good to go now.  Sorry about that!

kboone:
I downloaded the final version, and everything seems to work as expected. I'll merge this.

djones1040:
thanks so much!

PR #298, opened 2021-04-19
Update docs

kboone:
This contains a big update to the `Spectrum` docs, and bunch of small fixes to the rest of the docs for things that have broken over time.


docs/spectrum.rst
```diff
@@ -93,38 +89,45 @@ A spectrum can be rebinned with arbitrary wavelength bins. This returns a new
...
-example, to fit a single spectrum:
+Spectra can be used in fits. Any combination of spectra and photometry is
+allowed. To fit spectra, the times at which the spectra were taken must be
+specified. For example, to fit a single spectrum:
 
 .. code:: python
 
-    >>> model.set(z=0., amplitude=1., t0=0.)
-    >>> sncosmo.fit_lc(model=model, spectra=binned_spectrum,
+    Create the spectrum object, and specify the time at which it was taken.
```
benjaminrose, line 106: Should this line have a leading "#"?

benjaminrose: Same for L109.

benjaminrose:
Thanks @kboone! I'm appreciative that someone found time to work on the documentation.

Two related questions: 1) Once this is merged, should we start a release with SALT3 and these changes? 2) Is there something we can do to ensure future model additions don't break the "List of Built-in Sources"?

kboone:
To avoid this in the future, we should add a test to make sure that the docs compile without errors. There were errors in that build, but they weren't flagged when we were merging PRs.

Yes, I'll put out a new release when this is merged.

docs/spectrum.rst
kboone: Fixed in latest commit

PR #300, opened 2021-05-21
fixed bugs in the implementation of the SALT3 error model

djones1040:
sorry guys, accidentally had some bugs in SALT3Source _bandflux_rvar_single that was resulting in anomalously high chi^2 values from SALT3.  These should all be fixed now.

### DIFF
--- a/sncosmo/models.py
+++ b/sncosmo/models.py
@@ -1129,6 +1129,8 @@ class SALT3Source(SALT2Source):
         m0 = self._model['M0'](phase, wave)
         m1 = self._model['M1'](phase, wave)
         tmp = trans * wave
+
+        # evaluate avg M0 + x1*M1 across a bandpass
         f0 = np.sum(m0 * tmp, axis=1)/tmp.sum()
         m1int = np.sum(m1 * tmp, axis=1)/tmp.sum()
         ftot = f0 + x1 * m1int
@@ -1139,6 +1141,8 @@ class SALT3Source(SALT2Source):
         lcrv11 = self._model['LCRV11'](phase, band.wave_eff)[:, 0]
         lcrv01 = self._model['LCRV01'](phase, band.wave_eff)[:, 0]
 
+        # variance in M0 + x1*M1 at the effective wavelength
+        # of a bandpass
         v = (lcrv00 + 2.0 * x1 * lcrv01 + x1 * x1 * lcrv11)
 
         # v is supposed to be variance but can go negative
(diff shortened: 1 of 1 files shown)


sncosmo/models.py
```diff
@@ -1128,8 +1129,8 @@ def _bandflux_rvar_single(self, band, phase):
         m0 = self._model['M0'](phase, wave)
         m1 = self._model['M1'](phase, wave)
         tmp = trans * wave
-        f0 = np.sum(m0 * tmp, axis=1) * dwave / HC_ERG_AA
-        m1int = np.sum(m1 * tmp, axis=1) * dwave / HC_ERG_AA
+        f0 = np.sum(m0 * tmp, axis=1)/tmp.sum()
+        m1int = np.sum(m1 * tmp, axis=1)/tmp.sum()
```
kbarbary, line 1133: This seems like a big change from SALT2. Can you explain the logic and/or add a comment here?

kbarbary (commented):
Are there any tests you could add to ensure correctness? (I know it can be difficult to construct meaningful test cases for stuff like this).

sncosmo/models.py
djones1040: thanks Kyle, will do - this evaluates the average M0 + x1*M1 for a given bandpass.  The variance, "v" is the uncertainty in M0 + x1*M1 (in SALT2 it's a fractional error but here it's a true error).  So then to return the relative variance we divide v/ftot**2

sncosmo/models.py
```diff
@@ -1138,7 +1141,9 @@ def _bandflux_rvar_single(self, band, phase):
         lcrv11 = self._model['LCRV11'](phase, band.wave_eff)[:, 0]
         lcrv01 = self._model['LCRV01'](phase, band.wave_eff)[:, 0]
 
-        v = lcrv00 + 2.0 * x1 * lcrv01 + x1 * x1 * lcrv11
+        # variance in M0 + x1*M1 at the effective wavelength
+        # of a bandpass
+        v = (lcrv00 + 2.0 * x1 * lcrv01 + x1 * x1 * lcrv11)
```
kbarbary, line 1146: Style: Don't need the `( )` here.

kbarbary:
Anything else to do here? This seems important to fix!

kboone:
This all looks good to me, and I fixed the last style issue that @kbarbary brought up in #313 since I can't edit this PR. I'm going to go ahead and merge it. Thanks @djones1040!

djones1040:
hey guys, thank you so much - was just hiking and totally offline for the last three weeks, really appreciate you closing this out and please lmk if there's anything else I can do

Issue #308, opened 2021-08-18
SNCosmo installs a "tests" package


kboone:
The `sncosmo.tests` package gets installed as `tests`, so that when you do `import tests` it gets imported. I don't think that this behavior was intended, and it should probably only get imported as `import sncosmo.tests`.

kbarbary:
Yeah that was certainly not intended! I'm suprised that `packages=['sncosmo', 'sncosmo.tests']` in `setup()` results in this behavior. Anyway, is this fixed now with #309?

kboone:
Honestly, I'm not really sure how this happened. I tested it with a fresh install with #309 and it didn't happen, so it looks like it is fixed now.

Issue #304, opened 2021-08-03
Create precompiled wheels for sncosmo in pypi


kboone:
Currently we are uploading the sncosmo source to pypi, and it has to be built whenever it is downloaded. It would be better to produce compiled wheels like most other open source packages do. The compilation would need to happen in some CI tool, so this would be good to add when we get the GitHub Actions CI up and running.

kboone:
Fixed with #309 in v2.6.0

Issue #303, opened 2021-08-03
Broken pip install on Python 3.9


kboone:
Pip installation of sncosmo is currently broken on Python 3.9. The problem is that our current pyproject.toml file has a pinned numpy version 1.17.3 requirement for Python>=3.8, but Python 3.9 is only supported for numpy version 1.19.3 or above. This forces pip to rebuild numpy version 1.17.3 for Python 3.9 whenever sncosmo is installed, and this will fail on machines that don't have all of the numpy build dependencies installed.

With the current setup, this is going to be a recurring issue whenever there is a new Python release.

A better solution is to add a build dependency on [this scipy metapackage](https://github.com/scipy/oldest-supported-numpy/). Astropy is using this technique. That will automatically keep the build dependencies up to date. I'll make a PR for this.

Thanks to @jfcrenshaw for finding this.

kboone:
This bug also exists in the extinction package that sncosmo depends on, so both will need to be updated.

kboone:
This has now been fixed in both sncosmo and extinction, so this should be resolved.

Issue #245, opened 2019-06-10
SNCosmo 2.0 NumPy requirement is out of date - Import Error


djperrefort:
SNCosmo only requires `numpy >= 1.7.0`. However, when running NumPy version 1.15.4, importing SNCosmo 2.0 fails with the following error.

```
>>> import sncosmo
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/python3.7/site-packages/sncosmo/__init__.py", line 133, in <module>
    from .models import *
  File "/python3.7/site-packages/sncosmo/models.py", line 25, in <module>
    from .salt2utils import BicubicInterpolator, SALT2ColorLaw
  File "__init__.pxd", line 918, in init sncosmo.salt2utils
ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject

```

This error is not raised for the most recent version of Numpy 1.16.4.

kbarbary:
Did you install with conda? This might be an issue with how conda builds are specific to numpy runtime versions.

djperrefort:
No - I installed all packages using pip.

kbarbary:
Thanks that's helpful to know. I believe, but am not certain, that this is due to the new use of `pyproject.toml` to install build-time dependencies. Basically, at build time, pip is installing numpy and cython into an isolated environment, regardless of whether they are already installed on your system. It uses the most recent versions by default (numpy 1.16). Then sncosmo.salt2utils is build against that numpy (e.g., 1.16). But then at runtime, you might be using an older numpy, and get a failure.

I think the solution will be to  fix the version of numpy used in `pyproject.toml` to the oldest supported version. But I have to do some reading to check that this is really the best practices approach, and how old a version to use.

(I believe sncosmo is in fact compatible with numpy >= 1.7 in terms of the API. The binary compatibility is a different issue, but an important one!)

kbarbary:
Similar issue: astrofrog/fast-histogram#35

kbarbary:
Should be fixed in v2.1.0

benjaminrose:
@djperrefort, do you still have this error or should we close this issue as fixed?

kboone:
This should be permanently fixed with #305. We now rely on the scipy oldest-supported-numpy package to make sure that we are using compatible numpy versions for builds.

PR #309, opened 2021-08-18
Migrating to GitHub Actions and updating the build system

kboone:
This PR migrates the CI to GitHub Actions and updates the sncosmo build system. The current build system uses some older python conventions, and migrating to newer best practices would fix a bunch of the issues that have popped up recently. Namely, this should address all of the following issues: #245, #293, #303, #304, #308.

I started this PR off of #294 from @ykwang1. I setup everything by copying what is currently done by astropy. The main changes include:
- [x] Migrate away from `setup.py` to `setup.cfg`, and manage all dependencies in that file.
- [x] Implement all of the different CI tests in tox instead of having them tied to a specific CI implementation.
- [x] Set up GitHub Actions to run all of the tox tests.
- [x] Migrate to codecov instead of coveralls (this is what astropy uses, and it seems to be much more commonly used now).
- [x] Update oldest dependency requirements, and include them in the tests.
- [x] Automatically build wheels for sncosmo with cibuildwheel and publish them to PyPI when new versions are tagged.
- [x] Remove redundant configuration files.
- [x] Update all of the docs.

To test things, this PR is building the following:
- docs at https://kboone-sncosmo.readthedocs.io/en/latest/
- coverage at https://app.codecov.io/gh/kboone/sncosmo/branch/github_actions/
- wheels on TestPyPI at https://test.pypi.org/project/sncosmo/
- [x] Before merging the PR, make sure to point everything to the right location and adjust build frequencies.


kboone:
Several of the new build tools depend on Python 3.6. sncosmo currently claims to support Python 3.5, although we aren't currently testing it on Python 3.5 so I'm not sure if that is true. I propose bumping the minimum Python version up to 3.6 with this overhaul.

kboone:
This is now at a stage where all of the tests are passing in GitHub Actions including style, coverage (with codecov.io) and docs tests.

kboone:
To test things, this PR is building the following:
- docs at https://kboone-sncosmo.readthedocs.io/en/latest/
- coverage at https://app.codecov.io/gh/kboone/sncosmo/branch/github_actions/
- wheels on TestPyPI at https://test.pypi.org/project/sncosmo/

- [ ] Before merging the PR, I need to make sure to point everything to the right location and adjust build frequencies.

kboone:
Building wheels and uploading them to PyPI is now working properly, and I updated all of the documentation. I think that this is ready to go. @benjaminrose offered to do a review.

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+            os: ubuntu-latest
+            python: 3.6
+            toxenv: py36-oldestdeps
+
+          - name: Python 3.7 (MacOS X) with all optional dependencies
+            os: macos-latest
+            python: 3.7
+            toxenv: py37-alldeps
+
+          - name: Python 3.8 (Windows) with all optional dependencies
```
benjaminrose, line 50: Why is it python 3.8 on Windows but 3.7 on MacOS?

docs/contributing.rst
```diff
@@ -161,37 +171,75 @@ you can delete it by doing ``git branch -D simulation-enhancements``.
...
 ============================================
 
-These are notes mainly for the one person that manages releases.
-Yes, this could be more automated, but it isn't done very often,
-and involves some human verification.
+The release procedure is automated through GitHub Actions. To create a new
+release:
 
 - Update ``docs/history.rst`` with a summary of the new version's changes.
-- Bump version in ``sncosmo/__init__.py``.
```
benjaminrose, line 173: Is this not needed anymore? If not, when/where is this automated?

docs/contributing.rst
```diff
@@ -161,37 +171,75 @@ you can delete it by doing ``git branch -D simulation-enhancements``.
...
-- Build package and docs and check that docs look good.
-- Commit.
-- ``git clean -dfx``
-- ``tox`` (test in clean virtual env and make sure everything passes.
-- ``setup.py sdist upload`` (register new version and upload.)
-
-**Post-release steps:**
```
benjaminrose, line 180: Also, do we not need any new post-release prep for the next version?

setup.py
```diff
@@ -1,88 +1,7 @@
...
-          'Topic :: Scientific/Engineering :: Astronomy',
-          'Topic :: Scientific/Engineering :: Physics'],
+from extension_helpers import get_extensions
 
-      zip_safe=False,
-      use_2to3=False,
-      ext_modules=extensions,
-      packages=['sncosmo', 'sncosmo.tests'],
-      package_data={'sncosmo': data_files})
+setup(use_scm_version=True, ext_modules=get_extensions())
```
benjaminrose, line 7: What is `use_scm_version`?

sncosmo/__init__.py
```diff
@@ -8,7 +8,23 @@
...
+    from importlib.metadata import version, PackageNotFoundError
+    __version__ = version("sncosmo")
+except ImportError:
+    # Python 3.7 or lower. Use setuptools to find the version.
+    from pkg_resources import get_distribution, DistributionNotFound
+    try:
+        __version__ = get_distribution("sncosmo").version
+    except DistributionNotFound:
+        # Package is not installed
+        __version__ = '0.0.0'
```
benjaminrose, line 24: Is this the default version we want to use? I ask because I am not fully satisfied with '0.0.0' but I also don't have anything better.

Also, in what environments do expect to get these errors and set `__version__='0.0.0'`?

.github/workflows/run_tests.yml
```diff
@@ -0,0 +1,70 @@
...
+            toxenv: codestyle
+
+          - name: Build the docs
+            os: ubuntu-latest
+            python: 3.8
+            toxenv: build_docs
+
+          - name: Python 3.9 with minimal dependencies
+            os: ubuntu-latest
+            python: 3.9
```
benjaminrose, line 32: Should this always be the latest python version, so we don't need to update it annually? Can we do that with `python: 3.x`?

setup.cfg
```diff
@@ -0,0 +1,100 @@
...
+* = data/*, data/*/*, data/*/*/*, data/*/*/*/*, data/*/*/*/*/*, data/*/*/*/*/*/*
+sncosmo = sncosmo.cfg
+
+[coverage:run]
+source = sncosmo
+omit =
+   sncosmo/tests/*
+   sncosmo/conftest.py
+   sncosmo/__init__.py
+   sncosmo/version.py
```
benjaminrose, line 80: Are we still using `version.py`?

benjaminrose (commented):
Over all looks good. I don't fully understand the way we would be tracking sncosmo versions. I asked a few specific questions, but expect we can merge these changes soon.

docs/contributing.rst
benjaminrose: I think I am starting to understand once I read `__init__.py`. Maybe we should comment on how this is now automated, since there are many systems/tools to bump version numbers?

.github/workflows/run_tests.yml
kboone: I did that so that we would be running at least one test on each supported Python version (3.6, 3.7, 3.8, and 3.9). Astropy does the same thing.

docs/contributing.rst
kboone: The version is now managed by `setuptools_scm`. That package reads the git history and pulls the version from there directly. This approach seems to be the recommended one these days, and is used by astropy. With `setuptools_scm`, when you make a new release on github, it automatically picks it up and updates the version. Any development builds after that also automatically get tagged with the latest version and a dev string afterwards that includes the git commit.

The remaining challenge is how to handle the `__version__` label. There is an `importlib.metadata` package in the standard library in Python 3.8+ that can automatically pull the version from `setuptools_scm`, and this is the way of the future. However, we still want to maintain compatibility with Python 3.6/3.7. There are a few options for this:
1. There is a backport of `importlib_metadata` on PyPI that we could add as a dependency.
2. There is a function in `setuptools` that can handle this, but it then adds a runtime dependency on `setuptools` that adds a bit of extra latency. This is what I decided to do.
3. You can make `setuptools_scm` write out the version to a file on installation that can then be read at runtime. This is what `astropy` does, but it makes things much more complicated.

There are a lot of nice benefits to using the `setuptools_scm` package, and I think that it is the way of the future, but it is a little tricky to use it for Python 3.6/3.7.

I'm not opposed to dropping this and going back to the old hardcoded version if you think that's easier.

docs/contributing.rst
kboone: Nope, everything is automated now. As soon as you publish a new release on GitHub, everything else should happen automatically.

setup.py
kboone: That line enables `setuptools_scm`.

docs/contributing.rst
kboone: I guess that the one step that isn't automated is creating the feature branch (e.g. 2.5.x). We have never really done anything with those though, so I'm not convinced that they are necessary. I'd prefer to just stop making them.

sncosmo/__init__.py
kboone: I agree that this isn't a very good default. The environments that you get the errors from are ones that don't have the git history.

Playing around with different use cases a bit more, if you download a tarball of sncosmo from github `setuptools_scm` fails. It also completely prevents you from installing the package. This is also a problem for `astropy`! I'm not a big fan of that, so I'm thinking that reverting to automatically updating the version tag is the way to go.

.github/workflows/run_tests.yml
kboone: I think that we want to do this manually. GitHub might bump their runner up to the next version before our dependencies have fully migrated over leading to test failures that aren't the fault of sncosmo.

setup.cfg
kboone: After going through this review, I'm not convinced that `setuptools_scm` is making anything easier for us. I'm going to drop it and revert to the old method...

.github/workflows/run_tests.yml
benjaminrose: Seems good. Is there something we can do to try and notify us that we need to add a new test once a new python version comes out?

sncosmo/__init__.py
benjaminrose: Who downloads tarballs? Do we have a metric on how many people use that?

setup.cfg
benjaminrose: I laugh that you talked yourself into dropping `setuptools_scm` while simultaneously convincing me that we should use it.

benjaminrose:
A few questions that should be brought out of the code discussion and into the main thread:

Do we want to use `setuptools_scm` for version management? And related, do we have metrics on how many people use python 3.6/3.7 or tarballs?

docs/contributing.rst
benjaminrose: I like the idea of `setuptools_scm`. More comments in the main tread.

kboone:
`setuptools_scm` seems like a nice idea, and it is used by astropy and affiliated packages, but it has some tradeoffs. Namely, it breaks things like tarball installs and needs some tricks to work right on Python<3.8. It isn't necessary for this PR, so let's hardcode the version for now and figure this out later.

Similary, the `extension_helpers` tool that all of the astropy packages use is helpful, but it adds an extra dependency requirement if you want to run `python setup.py`. That is now deprecated anyway. The more modern `pip install` still works as expected, but we're losing some functionality for minimal benefit.

I'm going to remove both of those packages for this PR. We can discuss whether we want to include them in a separate PR. My vote is to just hardcode things for now and not worry about it.

benjaminrose:
Sounds good. One PR per idea is always good. Can you save a diff between a hard coded version and using `setuptools_scm` somewhere? That should help us later. Also, maybe open an issue with these comments:

> The version is now managed by `setuptools_scm`. That package reads the git history and pulls the version from there directly. This approach seems to be the recommended one these days, and is used by astropy. With `setuptools_scm`, when you make a new release on github, it automatically picks it up and updates the version. Any development builds after that also automatically get tagged with the latest version and a dev string afterwards that includes the git commit.
> 
> The remaining challenge is how to handle the `__version__` label. There is an `importlib.metadata` package in the standard library in Python 3.8+ that can automatically pull the version from `setuptools_scm`, and this is the way of the future. However, we still want to maintain compatibility with Python 3.6/3.7. There are a few options for this:
> 
> 1. There is a backport of `importlib_metadata` on PyPI that we could add as a dependency.
> 2. There is a function in `setuptools` that can handle this, but it then adds a runtime dependency on `setuptools` that adds a bit of extra latency. This is what I decided to do.
> 3. You can make `setuptools_scm` write out the version to a file on installation that can then be read at runtime. This is what `astropy` does, but it makes things much more complicated.
> 
> There are a lot of nice benefits to using the `setuptools_scm` package, and I think that it is the way of the future, but it is a little tricky to use it for Python 3.6/3.7.

kboone:
Latest version of the PR drops both `setuptools_scm` and `extension_helpers`. It will work fine even with tarballs, and can be built with the deprecated `python setup.py install` command as long as `numpy` is installed. I'll open an issue about those packages to discuss if we want to use them.

benjaminrose:
I think this is good to go then: tests pass, one change per PR, and it has been reviewed by a second set of eyes.

kboone:
We get Python 3.5 support back again if we drop those dependencies. I hope that nobody is using Python 3.5 since it is past EOL, but hey, might as well keep it in.

kboone:
I think that this is good to go now. All of the tests are passing (on my fork), the code coverage is syncing, and the docs are building. If there is no additional feedback, I'll merge this on Monday and start cleaning up the rest of the issues that I opened.

Issue #299, opened 2021-05-03
SNEMO model not fully documented


benjaminrose:
Several things:
- [x] No API documentation, unlike [SALT2](https://sncosmo.readthedocs.io/en/stable/api/sncosmo.SALT2Source.html)
- [x] Not apart of [Class Inheritance Diagram](https://sncosmo.readthedocs.io/en/stable/reference.html#class-inheritance-diagrams)

These issues are also present for other models. The source list is huge, so we likely don't want to add every model/source. Do we have an expectation for what should be a part of the API and class inheritance documentation?

Maybe we need to write a tutorial on how to add a new source, containing a checklist for PR reviews?

kbarbary:
While the list of built in sources is long, the number of unique _classes_ isn't that long: most of the sources are `TimeSeriesSource` instances and therefore all have the same API.

`SNEMO` should have been added to the docs when the class was added to sncosmo: that was an oversight! I believe fixing this is a two line change to https://github.com/sncosmo/sncosmo/blob/master/docs/reference.rst: add `SNEMO` to the list under "source components of model" and to the inheritance diagrams at the bottom of the page.

benjaminrose:
Thanks. #312 fixed this issue.

PR #312, opened 2021-09-02
Update the model documentation

kboone:
This addresses #299, and adds documentation for the SNEMO, SUGAR, and MLCS models.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -21,8 +21,11 @@ Model & Components
    Source
    TimeSeriesSource
    StretchSource
+   MLCS2k2Source
    SALT2Source
    SALT3Source
+   SNEMOSource
+   SUGARSource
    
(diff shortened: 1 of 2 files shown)


docs/reference.rst
```diff
@@ -21,8 +21,11 @@ Model & Components
    Source
    TimeSeriesSource
    StretchSource
+   MLCS2k2Source
    SALT2Source
    SALT3Source
+   SNEMOSource
+   SUGARSource
```
kbarbary, line 28: Can you also add these to the inheritance diagram on L145?

kboone: I updated the inheritance diagram to automatically document all classes in the models, magsystems, and bandpasses modules. That way we don't have to remember to add them in the future.

kbarbary (approved):

Issue #318, opened 2021-09-09
Improve cross-compatibility between SNANA and SNCosmo


benjaminrose:
I am working on fitting SNANA sims with SNIa models in SNCosmo. As a result, I am finding some unnecessary name conversions between the two programs. 

Between my `read_snana_fits()` and `mcmc_lc()` calls, I currently need to 
```python
data.rename_column("FLUXCAL", "flux")
data.rename_column("FLUXCALERR", "fluxerr")
data.rename_column("ZEROPT", "zp")
```

We can improve this workflow by either

1. adding these names to [`PHOTDATA_ALIASES`](https://github.com/sncosmo/sncosmo/blob/750865e13661f54b8e20b778e76b46b3f43a7561/sncosmo/photdata.py#L16)
2. or, adding an extra flag to `read_snana_fits()` that performs some standardization on names. 

I think option 1 is best, but option 2 could also add a `data["zpsys"] = "ab"` column if no zero-point system is found.

kboone:
The first option sounds good to me. Want to write up a quick PR and we can get it into v2.6?

Is the zeropoint in SNANA standardized internally, or are there models with a range of different zeropoints? We need to be very careful about assuming a zeropoint/zpsystem when one isn't specified.

benjaminrose:
I might be able to get a PR in, when are you releasing v2.6?

I am not sure if SNANA has a standard, but we can add an option where the user can provide the string that gets assigned to `data["zpsys"]`, but at that point it is probably better to just have the user specifically add the column after import.

kboone:
I'm hoping to release 2.6 today or tomorrow.

This change is just adding a couple of strings to a list. I can whip it up
in a couple mins if you want.

benjaminrose:
I'm working on it now. I just needed to finish another project first (and before the end of the day).

benjaminrose:
I am double-checking that `FLUXCAL` is what I think it is. In doing this, I found out that SNANA uses both `FLUX` and `FLUXCAL`. If a user imports a SNANA data file that has both, option 1 will cause a problem and break previously working scripts.

kboone:
Reading the SNANA manual, it looks to me like the `FLUX` column is the flux in ADU with the zeropoint given by `ZEROPT` while `FLUXCAL` is the flux scaled by 10^11 (a zeropoint of 27.5). So `FLUXCAL` is probably something that we need to handle separately.

[...]

PR #321, opened 2021-09-10
Add new zeropoint alias

benjaminrose:
Part of #318


kboone:
I think that this looks good.

Just to make sure that I understand this correctly, the `ZEROPT` flag should be used if we are reading the `FLUX` and `FLUXERR` columns from an SNANA light curve. It shouldn't be used if we are reading `FLUXCAL` and `FLUXCALERR`, but those keywords aren't in `photdata_aliases` so it should be fine. We need to add something to `read_snana_fits` (which will be a separate PR) to handle those. Is that right?

benjaminrose:
You are correct. If one uses FLUXCAL and FLUXCALERR then the ZP is always 27.5.

If we wanted to convert FLUXCAL to FLUX, that would be in `read_snana_fits()` but I am not sure we want to do that. Maybe, in another PR we could add a note in the documentation of `read_snana_fits()` stating the differences between FLUXCAL and FLUX and citing the SNANA manual.

PR #322, opened 2021-09-27
Add Roman WFI filters

benjaminrose:
A rewrite from PR #218.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -406,6 +406,23 @@ for name, fname in [('4shooter2::us', 'bandpasses/4shooter2/Us_4Shooter2.txt'),
     _BANDPASSES.register_loader(name, load_bandpass_remote_aa,
                                 args=(fname,), meta=fourshooter_meta)
 
+roman_meta = {
+     'filterset': 'roman-wfi',
+     'dataurl': 'https://roman.gsfc.nasa.gov/science/WFI_technical.html',
+     'retrieved': '30 Nov 2020',
+     'description': 'Roamn filters from Jeff Kruk: '
+     'Roman_effarea_20201130.txt '}
+for name, fname in [('f062', 'roman_f063dat'),   # R
+                    ('f087', 'roman_f087.dat'),  # Z
+                    ('f106', 'roman_f106.dat'),  # Y
+                    ('f129', 'roman_f129.dat'),  # J
+                    ('f158', 'roman_f158.dat'),  # H
+                    ('f184', 'roman_f184.dat'),  # F
+                    ('f213', 'roman_f213.dat'),  # K
+                    ('f149', 'roman_f149.dat')]:  # Wide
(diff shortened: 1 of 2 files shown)


sncosmo/builtins.py
```diff
@@ -406,6 +406,23 @@ def tophat_bandpass_um(ctr, width, name=None):
     _BANDPASSES.register_loader(name, load_bandpass_remote_aa,
                                 args=(fname,), meta=fourshooter_meta)
 
+roman_meta = {
+     'filterset': 'roman-wfi',
+     'dataurl': 'https://roman.gsfc.nasa.gov/science/WFI_technical.html',
+     'retrieved': '30 Nov 2020',
+     'description': 'Roamn filters from Jeff Kruk: '
+     'Roman_effarea_20201130.txt '}
+for name, fname in [('f062', 'roman_f063dat'),   # R
```
kboone, line 415: Looks like you missed a dot here.

sncosmo/builtins.py
```diff
@@ -406,6 +406,23 @@ def tophat_bandpass_um(ctr, width, name=None):
...
+                    ('f087', 'roman_f087.dat'),  # Z
+                    ('f106', 'roman_f106.dat'),  # Y
+                    ('f129', 'roman_f129.dat'),  # J
+                    ('f158', 'roman_f158.dat'),  # H
+                    ('f184', 'roman_f184.dat'),  # F
+                    ('f213', 'roman_f213.dat'),  # K
+                    ('f149', 'roman_f149.dat')]:  # Wide
```
kboone, line 422: Looks like this is named f146 on the sncosmo.github.io website?

sncosmo/builtins.py
kboone: Also, this is f062 on sncosmo.github.io, not f063.

sncosmo/builtins.py
```diff
@@ -406,6 +406,23 @@ def tophat_bandpass_um(ctr, width, name=None):
     _BANDPASSES.register_loader(name, load_bandpass_remote_aa,
                                 args=(fname,), meta=fourshooter_meta)
 
+roman_meta = {
+     'filterset': 'roman-wfi',
+     'dataurl': 'https://roman.gsfc.nasa.gov/science/WFI_technical.html',
+     'retrieved': '30 Nov 2020',
+     'description': 'Roamn filters from Jeff Kruk: '
```
kboone, line 413: Roamn -> Roman

kboone (changes requested):
There are a couple of typos and mismatches between the names on sncosmo.github.io and in the builtins.py, but other than that this should be good to go.

Also I noticed that your name in commit messages is set to "Benjain Rose". Not a big deal, but might want to fix that.

benjaminrose:
Somehow missed the first messages. I can fix these tomorrow.

Issue #323, opened 2021-10-08
Add tests for all of the builtins.


kboone:
It could be good to add a test to make sure that all of the builtin bandpasses and models are working/accessible.

We test a small number of them here right now: https://github.com/sncosmo/sncosmo/blob/master/sncosmo/tests/test_builtins.py

Instead of adding new functions for each bandpass, we should add something that just tests all of the builtin bandpasses and models automatically.

Some potential issues here are that this might be very slow to run due to the large number of downloads, and that downloads can fail stochastically. This should be benchmarked/investigated. We might want to run these tests in a separate tox environment from the main tests.

PR #324, opened 2021-10-12
Add tests for all builtins

kboone:
This addresses #323 and adds tests to make sure that all of the builtins work and are accessible.

Downloading all of the builtins takes a few mins and downloads 1.2 GB. This isn't something that needs to happen in every test run locally, so I made a separate tox environment for it. To test the builtins, run `tox -e builtins`. The tox environment runs everything in a clean environment and doesn't used cached data to make sure that all files can be downloaded. I added a new github action for this.

I also added new tests to `test_builtins.py` to test each of the different loaders that are used by the builtins. These will use the cache if available and are included in the standard tests.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -199,6 +199,12 @@ Running the tests with the ``-cov`` tag will generate a coverage report::
 
       tox -e py3-cov
 
+SNCosmo includes hundreds of builtin bandpasses and sources that are downloaded
+from external sites when they are loaded. ``tox`` can be used to check that all
+of these builtins are accessible with the following command::
+
+      tox -e builtins
+
 ``tox`` can also be used to check the code style::
 
       tox -e codestyle


kboone:
> Just so I understand what is intended: Is `test_builtins.py` being used as a spot check and ran along side the main set of tests, and with all supported python versions? And `test_download_builtins.py` is the full set of tests but will only be ran once?

Yes. `test_builtins.py` is testing all of the code in `builtins.py`. There are all sorts of functions like `load_bandpass_remote_aa`, and there is one test for each of those. I'm just loading one object for each of those functions. This also uses the sncosmo cache if it is available so if you are running it locally it won't do any downloads after the first time. It should be fast because there are only ~20 tests, and we want to run this on everything to catch bugs like the Windows one that I found when writing these tests.

`test_download_builtins.py` tests that all of the builtins work from a fresh installation. It uses a clean cache which forces it to download everything. There are several hundred builtins so this takes a few mins. This is really just making sure that all of the URLs are accessible and that things are in the right format, so I don't think that it needs to be run on every platform.

benjaminrose:
Thanks. I was reading things correctly. I think we should add a one line purpose comment to each test file. That way in a year, we will remember what type of test of `builtin.py` goes in what file. The file names are good, but a sentence will remove all possible confusion.

benjaminrose:
> I think we should add a one line purpose comment to each test file. 

Sorry, I now see that this is already in `test_download_builtins.py` but we should also add it to `test_builtins.py`.

PR #326, opened 2021-10-13
Update codecov.io uploader version

kboone:
I got an email that codecov.io has deprecated the uploader that we are using. See https://github.com/codecov/codecov-action for details. This PR changes our github action to use the new version.


kboone:
The codecov failure here isn't real. Because of the deprecation, codecov is doing a brownout and didn't run on the latest commit (3c7a6e7). It is comparing to 1a6e22d instead that doesn't contain the updated codecov configuration from 5c488a8. What a mess.

Merging this will avoid the brownout and things should go back to normal after.

Issue #328, opened 2021-10-14
Interpolating built-in source snf-2011fe results in partially negative spectrum


JTerwel:
I stumbled upon this while playing around with the built-in TimeSeriesSource for SN2011fe.
Interpolating the spectrum to around 40 days after peak seems to cause the blue part of the spectrum to go negative. As a result of this, when requesting an interpolated light curve in e.g. ztfg there is a region where light curve disappears.

For instance, by running the following code:
```
import sncosmo
import matplotlib.pyplot as plt
import numpy as np

x = np.linspace(-15, 95, 111)
test = sncosmo.get_source('snf-2011fe')
plt.plot(test._phase, test.bandmag('ztfg', 'ab', test._phase), label='built-in')
plt.plot(x, test.bandmag('ztfg', 'ab', x), label='built-in interpolated')
plt.xlim(-20, 110)
plt.ylim(16, 9.5)
plt.legend(loc='upper right')
plt.show()

plt.plot(test._wave, test.flux(40, test._wave))
plt.show()
```
I plot the light curve, while only using the phases of the spectra in the TimeSeriesSource (No interpolation, blue curve).
I also plot the light curve by interpolating to get the magnitude each day (orange curve).

The second plot shows the interpolated spectrum at phase=40. The blue part of the spectrum, up to 5500 Angstrom is too low.

This is the result from the code above:
<img width="978" alt="Screenshot 2021-10-14 at 15 39 40" src="https://user-images.githubusercontent.com/82818204/137328980-bf56e9e0-d3d0-4127-9482-1e2c54ca5157.png">

The issue seems to come from the spline interpolation combined with the large gap between spectra in the TimeSeriesSource (last one is at phase=23.7 & next one is at phase=74.1) allowing inaccuracies in the interpolation to grow.
It might be possible to remedy this by changing the order of the spline, but I have yet to find a way to change it. kx & ky are currently both fixed at 3 inside the TimeSeriesSource.

So in short, I found an interpolation problem within the built-in SN2011fe source. The problem itself might however be more general if it is rooted in the spline interpolation in a TimeSeriesSource.

kboone:
Yes, this is a problem that shows up for several of the sparsely sampled TimeSeriesSources. There was discussion about this in #203 from a few years ago.

We should probably be using linear interpolation for time series like this one with large gaps instead of spline interpolation.

kboone:
[...]

Issue #203, opened 2017-05-21
nugent sn iip model has extremely large interpolation errors in sncosmo


dannygoldstein:
Because Peter's SNIIP template is defined on the following non-uniform time grid:

0.0
1.0
6.0
11.0
16.0
21.0
36.0
61.0
86.0
111.0
136.0
411.0

sncosmo's cubic splines cause a very large, spurious feature in the light curves between 136 and 411 days. 

![image](https://cloud.githubusercontent.com/assets/2769632/26281424/c1a0d074-3daa-11e7-934a-9035f9c80eb6.png)

This could probably be fixed by just ignoring the day 411 point in the SED.

dannygoldstein:
Even doing that, it's still pretty bad, the splines make the light curves too wiggly.

![image](https://cloud.githubusercontent.com/assets/2769632/26281451/8d05edd0-3dab-11e7-820a-333faacfcc26.png)

dannygoldstein:
Linear interpolation is best:

![image](https://cloud.githubusercontent.com/assets/2769632/26281460/f63147b4-3dab-11e7-93a0-eff06f2886b9.png)

rbiswas4:
Wonder how important this is for the other core-collapse templates!

kbarbary:
Yeah, should fix this.... Probably by adding a spine order option to
TimeSeriesSource and using reasonable settings for built in sources.

Did you try quadratic? Just curious.

dannygoldstein:
Yeah it's also pretty bad.

PR #329, opened 2021-10-14
Allow the spline degree in TimeSeriesSource to vary

kboone:
Many of the time series that are used for builtin models are quite sparse. This can be problematic for the cubic interpolation that is done in TimeSeriesSource. See e.g. #203 and #328.

This PR adds an option for setting the spline degree in TimeSeriesSource. For problematic models such as SN2011fe, we can then use linear interpolation rather than cubic interpolation which is more stable.

### DIFF
--- a/sncosmo/models.py
+++ b/sncosmo/models.py
@@ -527,7 +527,8 @@ class TimeSeriesSource(Source):
         self._phase = phase
         self._wave = wave
         self._parameters = np.array([1.])
-        self._model_flux = Spline2d(phase, wave, flux, kx=time_spline_degree)
+        self._model_flux = Spline2d(phase, wave, flux, kx=time_spline_degree,
+                                    ky=3)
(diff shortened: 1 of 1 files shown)


kboone:
Doing some quick tests, looks like the nugent-hyper, nugent-sn2p, nugent-sn2l, and nugent-sn2n models all have undesirable behavior with the cubic interpolation and could be improved with linear interpolation.

JTerwel:
With this change you can choose the spline degree in TimeSeriesSource, but the degree in both directions (wavelength & time) is forced to be the same. Is there a reason for this? If the coverage in time is sparse but good in wavelength, would it make sense to do linear interpolation in the time direction only while keeping it cubic in the wavelength direction?

I'm not super familiar with spline interpolation, so I don't know if is desirable to have the spline degree the same in both directions.

PR #336, opened 2021-12-28
Add ATLAS as a builtin

mcoughlin:
Given that ATLAS provides a very popular forced photometry service used by transient folks, having its filters would seem appropriate.

It looks like someone has to add the SVO files to http://sncosmo.github.io/data, but otherwise seems straightforward.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -472,6 +472,19 @@ for filter_id in ['open', 'g', 'r', 'i', 'z', 'y', 'w']:
                                 args=(fname,),
                                 meta=ps1_meta)
 
+# ATLAS
+atlas_meta = {
+    'filterset': 'atlas',
+    'retrieved': '28 Dec 2021',
+    'dataurl': ('http://svo2.cab.inta-csic.es/svo/theory/fps/getdata.php?'
+                'format=ascii&id=Misc'),
+    'description': 'ATLAS filters from SVO (includes filter, optics, detector and atmosphere.)'}
+for filt in ['Cyan', 'Orange']:
+    name = 'atlas' + filt[0].lower()
+    relpath = 'bandpasses/atlas/Atlas.{}'.format(filt)
+    _BANDPASSES.register_loader(name, load_bandpass_remote_aa,
+                                args=(relpath,), meta=atlas_meta)
+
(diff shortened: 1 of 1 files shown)


mcoughlin:
@benjaminrose I fixed the line too long I hope.

benjaminrose:
It looks like you are missing a closing `)` in the last commit. But more importantly is that download [failed](https://github.com/sncosmo/sncosmo/runs/4748731241?check_suite_focus=true). Are the filters defined somewhere, or do they need to be uploaded to sncosmo.github.io & maintained by SNCosmo?

mcoughlin:
They are here: http://svo2.cab.inta-csic.es/svo/theory/fps/index.php?id=Misc/Atlas.orange&&mode=browse&gname=Misc&gname2=Atlas

So if they can be pulled over to sncosmo.github.io would be great.

benjaminrose:
I think I added the needed redirects to the website, https://github.com/sncosmo/sncosmo.github.io/commit/d4c4dca8b7f6c7aadc6b47f48e2be3bff1176ad4.

mcoughlin:
@benjaminrose And I think I got that parentheses.

mcoughlin:
@benjaminrose maybe this test just needs to be retriggered?

benjaminrose:
New builtins need an update to both the code and the website. I tried to update the website in a slightly different, but I hoped more sustainable, way. It seems to not have worked. I’ll update it the old way soon and then the last test should pass. 

Once we get that passing, we should be able to merge this PR.

mcoughlin:
Thanks @benjaminrose !

mcoughlin:
@benjaminrose Is there anything I can to support getting the tests to pass?

kboone:
It looks like the last failure was a stochastic one and unrelated to this PR. I'm rerunning the tests now.

kboone:
LGTM

mcoughlin:
Thanks @benjaminrose and @kboone !

benjaminrose:
Thanks @kboone for rerunning it. And I am glad that we can use `redirects.json` rather than hosting the throughputs ourselves.

@mcoughlin, sorry for taking a while to get back to this PR. Glad it got the attention it needed.

PR #347, opened 2022-08-17
Use assert_allclose for comparison in test_salt2colorlaw_vs_python

olebole:
On Debian, the package is built and tested on several platforms with different CPU architectures.

On some architectures (i386, arm64, ppc64, riscv64), [one test fails](https://buildd.debian.org/status/fetch.php?pkg=sncosmo&arch=arm64&ver=2.8.0-1&stamp=1660665749&raw=0) because it compares two arrays computed in different ways to be equal. The assert fails here, since on these platforms the result is slightly (negligible) different.

The PR replaces the "=" assert with np.testing.assert_allclose, which allows small deviations. This [fixes this failure](https://buildd.debian.org/status/fetch.php?pkg=sncosmo&arch=arm64&ver=2.8.0-2&stamp=1660686860&raw=0).

### DIFF
--- a/sncosmo/tests/test_salt2utils.py
+++ b/sncosmo/tests/test_salt2utils.py
@@ -104,7 +104,7 @@ def test_salt2colorlaw_vs_python():
     colorlaw = SALT2ColorLaw(colorlaw_range, colorlaw_coeffs)
 
     wave = np.linspace(2000., 9200., 201)
-    assert np.all(colorlaw(wave) == colorlaw_python(wave))
+    np.testing.assert_allclose(colorlaw(wave), colorlaw_python(wave))
 
 
 def test_salt2colorlaw_pickle():


benjaminrose:
Are the defaults `rtol` and `atol` in `assert_allclose` the right values? What level of  discrepancy is seen on these architectures?

olebole:
I am not sure; the defaults worked for me. I would just trust them, unless you really have a real idea how accurate the result *should* be.

benjaminrose:
@olebole, do you need these patches in a released version before you can finalize your Debian work?

olebole:
I can always apply patches locally. Usually I mainly forward the patches (and bugs) which I consider as upstream problems. Having them applied upstream at some point helps me maintaining, but a new version is not necessary for me.

benjaminrose:
And we are very grateful for your upstream patches.

olebole:
FYI, a few links. I'll report when problems appear but this is if you want to browse yourself:
 * the Debian maintainer's page for sncosmo: https://tracker.debian.org/pkg/sncosmo
 * our git repository (indepenent of yours): https://salsa.debian.org/debian-astro-team/sncosmo
 * therein, the list of patches we apply: https://salsa.debian.org/debian-astro-team/sncosmo/-/tree/master/debian/patches
 * the Debian bug tracker for sncosmo: https://bugs.debian.org/cgi-bin/pkgreport.cgi?repeatmerged=no&src=sncosmo
 * all build logs: https://buildd.debian.org/status/logs.php?pkg=sncosmo

PR #357, opened 2023-03-08
TESS

mcoughlin:
This PR adds TESS to builtin filters (relevant given its popularity and public data). Pairs with https://github.com/sncosmo/sncosmo.github.io/pull/32.

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -523,7 +523,7 @@ tess_meta = {
     'description': 'TESS filter from SVO (includes filter and instrument)'
 }                
 for filt in ['Red']:
-    name = 'tess::' + filt.lower()
+    name = 'tess'
     relpath = 'bandpasses/tess/tess.{}'.format(filt)
     _BANDPASSES.register_loader(name, load_bandpass_remote_aa,
                                 args=(relpath,), meta=tess_meta)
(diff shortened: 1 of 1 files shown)


mcoughlin:
@benjaminrose

mcoughlin:
@benjaminrose Anything I can do to help? I added you as a collaborator on my sncosmo.github.io branch as it looks like you want a different name for the TESS filter.

benjaminrose:
There are a lot of failures, but they don't look like they have to do with the TESS filters, except for "Download all builtins." But that URL worked for me just now, so my guess is that `HTTP Error 502` is a internet issue and not a SNCosmo issue, especially since the URL worked just now.

mcoughlin:
Thank you @benjaminrose. The last thing is to understand when the next release might be possible.

benjaminrose:
We usually wait for a few new things per release. I especially want to do that with all the issues we have in our CI system. Is it possible for you to [use the GitHub](https://matiascodesal.com/blog/how-use-git-repository-pip-dependency/) version for now?

mcoughlin:
@benjaminrose Of course, I just try to avoid such dependencies when I can. Thanks for getting this in.

mcoughlin:
Thanks @benjaminrose!

PR #374, opened 2023-06-16
GOTO

mcoughlin:
This PR adds GOTO as a builtin bandpass (important given they are publicly announcing candidates in their filter set).

### DIFF
--- a/sncosmo/builtins.py
+++ b/sncosmo/builtins.py
@@ -530,6 +530,20 @@ for filt in ['Red']:
                                 args=(relpath,), meta=tess_meta)
 
 
+# GOTO
+goto_meta = {
+    'filterset': 'goto',
+    'retrieved': '16 June 2023',
+    'dataurl': ('http://svo2.cab.inta-csic.es/svo/theory/fps/getdata.php?'
+                'format=ascii&id=Misc'),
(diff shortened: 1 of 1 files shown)


mcoughlin:
@benjaminrose

mcoughlin:
Thanks for merging @benjaminrose. I hope this one is now good to go.

benjaminrose:
Yeah, sorry for the delay. My one question: GOTO is not a common telescope/acronym, I wonder if we should use the full name for `'filterset': 'The Gravitational-wave Optical Transient Observer'`. Thoughts?

mcoughlin:
@benjaminrose could do, although then shouldn't we have done that for atlas / panstarrs/ ztf and everything else?

benjaminrose:
Yes, this matches what is done elsewhere, so is viable. But Atlas, panSTARRS and ZTF are all very common surveys. Let's leave it as is, and people can go to `svo2.cab.inta-csic.es/svo/theory/fps` and get the details about the filters.

Again, sorry for being a bit slow. I was traveling last week and the second PR slipped through the cracks.

mcoughlin:
Thanks @benjaminrose!

Issue #373, opened 2023-06-12
Add SNIa intrinsic scatter functions


benjaminrose:
If we want to do this, it would allow for something like
```
obs_fluxes = SALT2_model(x1, c, ...) + scatter.g10(x1, c) + measurement_noise
```

I want to have `grey`, `g10`, `c11` and `bs21` models.

benjaminrose:
Maybe make this an collection of effects so they stay with the source object and can automatically be used by the simulation utility in SNCosmo.

PR #378, opened 2023-07-12
adding G10 and C11 intrinsic scattering models

bastiencarreres:
Hi, I propose this implementation for the G10 scattering model, it's a cleaner implementation of one I use in [snsim](https://github.com/bastiencarreres/snsim). It follows the implementation used in SNANA as described in [arxiv:1209.2482](https://arxiv.org/pdf/1209.2482.pdf). 

I create the `G10` class in `models.py` that inherit from `PropagationEffect`. To be initialized the class need a `SALTsource`:
```
import sncosmo as snc

SALTSource = snc.models.get_source(name='salt2')
G10 = snc.models.G10(Source)
SALTwithG10 = snc.Model('salt2')
SALTwithG10.add_effect(G10, name='G10', frame='rest')
```
The `compute_sigma_nodes` class function compute a sigma value from `source.minwave()` to `source.maxwave()` each 800 A, then for each sigma an intrinsic scattering is drawn with gaussian pdf. The result is two arrays : the wavelength at values  `lam_nodes` and the scattering value for each wavelength `siglam_values`. 

In `propagate` class function the wavelengths are interpolated by a sinus function on  (`lam_nodes` , `siglam_values`).

Possible problem: The Gaussian drawing is done each time `propagate` is called, resulting in different results for different calls of `SALTwithG10.bandflux`.


MickaelRigault:
I would like to comment, that I like that very much. 
We should also have others like C11 etc. and an example on how to build your own.
Thanks @bastiencarreres

bastiencarreres:
Hi @MickaelRigault, I also have the C11 but need some extra-test before push I think

bastiencarreres:
I modfied the PR, I've added the `C11` class in `model.py`.
```
import sncosmo as snc

Source = snc.models.get_source(name='salt3')
C11 = snc.models.C11()
ModelC11 = snc.Model('salt3')
ModelC11.add_effect(C11, name='C11',frame='rest')
```

I followed the SNANA implementation also described in [arxiv:1209.2482](https://arxiv.org/abs/1209.2482).
I added a `sine_interp(x_new, fun_x, fun_y)` function in `utils.py`. This function is used by G10 and C11.

The implementation of C11 works as follows:
*There is 2 parameters inside `self._parameters` : the first one is `C_vU` is the correlation between the v and U bands and is usually -1, 0 or 1(maybe need to add an error on this); the second parameter `S_f` is a scaling factor for the covariance fixed to 1.3 in  [arxiv:1209.2482](https://arxiv.org/abs/1209.2482).
*The `self._lam_nodes` are defined as the central wavelengths of vUBVRI bands.
*The correlation matrix from Chotard et al. 2011 between vUBVRI bands `self._corr_matrix` is converted into a covariance `self._cov_matrix` matrix by multiplying by sigma values `self._siglam_values`. 
*The correlation matrix `self._cov_matrix` is then rescale by the scaling factor `S_f`
*When `propagate(self, wave, flux)` is called `siglam_values` are drawn on a multivariate normal centered on 0 and with the defined covariance matrix, then the `wave` are interpolated with `sine_interp` on (`self.lam_nodes`, `siglam_values`), wavelengths outside the vUBVRI range are interpolated as the `siglam_values[0]` or  `siglam_values[-1]` constant.

As for G10 the Gaussian drawing is done each time propagate is called, giving different results at each call.

MickaelRigault:
@benjaminrose 

Any update on this ? Could that be accepted or you think it should not be made like that ?

benjaminrose:
After a quick review, `SALTwithG10.bandflux` should be repeatable and not vary by call. I would like to see an example in the docs. Either in the [model & effects section](https://sncosmo.readthedocs.io/en/stable/models.html#phase-dependant-effects) or in the [simulation section](https://sncosmo.readthedocs.io/en/stable/simulation.html) (or both?). Also I want to run the test suite. Can you add a few tests similar to the ones found here
 https://github.com/sncosmo/sncosmo/blob/903706066adb390d45e33655e466fca402d340e0/sncosmo/tests/test_models.py#L342-L369.

sncosmo/models.py
```diff
@@ -2030,3 +2030,95 @@ def propagate(self, wave, flux, phase=None):
...
+        
+        siglam_values[lam_nodes < L0] *= 1 + (lam_nodes[lam_nodes < L0] - L0) * F0
+        siglam_values[lam_nodes > L0] *= 1 + (lam_nodes[lam_nodes > L0] - L0) * F1
+        siglam_values *= np.random.normal(size=len(lam_nodes))
+        
+        return lam_nodes, siglam_values
+
+    def propagate(self, wave, flux):
+        """Propagate the effect to the flux."""
+        lam_nodes, siglam_values = self.compute_sigma_nodes()
```
benjaminrose, line 2066: Would moving this to `__init__` make this `SALTwithG10.bandflux` the same on repeated calls?

sncosmo/models.py
```diff
@@ -2030,3 +2030,95 @@ def propagate(self, wave, flux, phase=None):
...
+        # Rescale covariance as in arXiv:1209.2482
+        self._cov_matrix *= self._parameters[1]
+    
+    def propagate(self, wave, flux):
+        """Propagate the effect to the flux."""
+        siglam_values = np.random.multivariate_normal(np.zeros(len(self._lam_nodes)), self._cov_matrix)
+        
+        inf_mask = wave <= self._lam_nodes[0]
+        sup_mask = wave >= self._lam_nodes[-1]
+
```
benjaminrose, line 2118: I think these first three statements could be moved to `__init__`. Am I missing something?

sncosmo/models.py
bastiencarreres: Yes it should works since the random part is in `self.compute_sigma_nodes()`

bastiencarreres:
With respect of @benjaminrose comments, I moved all random variables in `__init__`.

bastiencarreres:
I'va added tests for G10 and C11 in `test_models.py` as two test function `test_G10` and `test_C11`. I've also done some debugging relative to the previous change that prevented  `Model.set()` from working.

sncosmo/models.py
bastiencarreres: Moving this part in `__init__ `  leads to a non-functional `Model.set()` since the updates on `self._parameters` will not be taken into account. I set a `self._seed` parameter in the `__init__` to fix the randomness when init the model.

benjaminrose:
#373

docs/models.rst
```diff
@@ -373,4 +414,4 @@ parameters:
     >>> source = sncosmo.SALT2Source(modeldir='/path/to/dir',
     ...                              m0file='mytemplate0file.dat')
 
-See `~sncosmo.SALT2Source` for more details.
+See  for more details.
```
benjaminrose, line 417: Why did this change?

docs/models.rst
```diff
@@ -246,6 +246,47 @@ this value is perfectly known from the dust map. Therefore, when using
...
+                                    effects=[G10],
+                                    effect_names=['G10'],
+                                    effect_frames=['rest'])
+
+The G10 model parameters are:   
+
+* ``L0``, ``F0`` and ``F1`` are used in the multiplicative factor introduced in Kessler et al. 2012. Their default values are ``L0=2157.3``, ``F0=0`` and ``F1=1.08e-4``.
+* ``dL`` the wavelength range between each scatter node. Following Kessler et al. 2012 it is set by default to 800A.
+
+The C11 model can be added to your `~sncosmo.Model` as:
```
benjaminrose, line 273: Is there an easy way to clearly state that C11 does not need a `G10 = snc.models.G10(SALTsource=SALTSource)` step?

docs/models.rst
```diff
@@ -246,6 +246,47 @@ this value is perfectly known from the dust map. Therefore, when using
...
+by G10 or C11 models. The implemention is based on Kessler et al. 2012.
+They both act as random variation in the spectra model of a `~sncosmo.SALT2Source` or `~sncosmo.SALT3Source`.
+
+The G10 model take a `SALTSource` as an argument and can be added to your `~sncosmo.Model` as:
+
+.. code:: python
+
+    >>> source = 'salt2'
+    >>> SALTSource = sncosmo.models.get_source(name=source)
+    >>> G10 = snc.models.G10(SALTsource=SALTSource)
```
benjaminrose, line 262: This document currently assumes `import sncosmo` rather than `import sncosmo as snc`.

docs/models.rst
```diff
@@ -246,6 +246,47 @@ this value is perfectly known from the dust map. Therefore, when using
...
+The G10 model parameters are:   
+
+* ``L0``, ``F0`` and ``F1`` are used in the multiplicative factor introduced in Kessler et al. 2012. Their default values are ``L0=2157.3``, ``F0=0`` and ``F1=1.08e-4``.
+* ``dL`` the wavelength range between each scatter node. Following Kessler et al. 2012 it is set by default to 800A.
+
+The C11 model can be added to your `~sncosmo.Model` as:
+
+.. code:: python
+
+    >>> C11 = snc.models.C11()
```
benjaminrose, line 277: Please change `snc` to `sncosmo`.

benjaminrose (commented):
Sorry this is taking a long time. This is a fairly major change in functionality and I want to use it as a template for future additions. I think we are really close to something extremely useful.

docs/models.rst
bastiencarreres: Sorry, probably done a ctr+X instead of ctrl+C. Fixed.

docs/models.rst
bastiencarreres: Fixed !

benjaminrose:
There are two issues found in the tests. 1) the tests don't work on python 3.7 and we should discuss what to do to fix that. and 2) there are some code style issues.

bastiencarreres:
I've corrected the code style. For python 3.7 I think the problem comes from `np.random.SeedSequence()`. A way to avoid that is to use the standard library function `secrets.randbits` as shown [here](https://numpy.org/doc/stable/reference/random/index.html).

benjaminrose:
Python 3.7 is also end of life, https://devguide.python.org/versions/, maybe we should just bump our minimum dependancies?

bastiencarreres:
I've re-corrected style. Yes Python 3.7 is end of life and not supported by numpy since 1.22 so changing minimum dependencies is a good solution !

benjaminrose:
Let me drop 3.7 support, then we should be able to merge this in.

MickaelRigault:
Hello guys,

Any update on this?

PR #385, opened 2024-01-05
MNT: upgrade upload/download-artifact GitHub actions

neutrinoceros:
Hi, version 4 of upload-artifact and download-artifact were recently released.
Staying on old version is not viable long term because GitHub will sooner or later drop support for the old versions of node that these depend on.
Since migration to v4 is not as straightfoward as bumping the version number, I'm helping all astropy-related packages byt manually performing the necessary changes.

### DIFF
--- a/.github/workflows/upload_to_pypi.yml
+++ b/.github/workflows/upload_to_pypi.yml
@@ -27,8 +27,9 @@ jobs:
         continue-on-error: true
         uses: pypa/cibuildwheel@v2.1.1
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


benjaminrose:
Tests are passing, except on Python 3.12. I'll merge this, then fix the distutil issue [#387].

benjaminrose:
There was an issue with downloading the artifacts. The original method put the wheels into subfolders inside `dist/`. Commit 52c749de puts the wheel content directly into `dist/`. This allows the pypi upload to use `twine check` successfully.

neutrinoceros:
Of course, sorry I missed that, and thanks for letting me know !

PR #386, opened 2024-01-12
Update on the DATAMIRROR, using secure http now

BrunoSanchez:
I had the issue that a new laptop, new fresh install, won't download the models using this data mirror just because of the http instead of https in the url.

Didn't want to open an issue for such a minor change.


benjaminrose:
Tests are passing, except on Python 3.12. I'll merge this, then fix the distutil issue [#387].

PR #389, opened 2024-02-08
Salt2 ColorLaw function correction

ConstanceGanot:
Hi, an error has been found when Pickling/Unpickling objects than contain sncosmo.models _Model_ objects : it updates the _SALT2ColorLaw_ object coefficients (in salt2utils.pyx) and then crash. 

It is due to the function _\_\_get_new_args\_\__ that has been made for pickling,  and that adds an extra value in the coefficients array to return.
I show an illustration of the bug in the picture below, we begin with some coefficients for the colorlaw and then it add 0s. 

I replaced **ncoeffs**, that is be the number of input coefficients +1,  to **ncoeffs-1** in the _\_\_get_new_args\_\__ function.

bug_colorlaw

### DIFF
--- a/sncosmo/salt2utils.pyx
+++ b/sncosmo/salt2utils.pyx
@@ -482,6 +482,6 @@ cdef class SALT2ColorLaw(object):
         # reconstruct input wavelengths
         wave_lo = self.l_lo * SALT2CL_V_MINUS_B + SALT2CL_B
         wave_hi = self.l_hi * SALT2CL_V_MINUS_B + SALT2CL_B
-        coeffs = [self.coeffs[i+1] for i in range(self.ncoeffs)]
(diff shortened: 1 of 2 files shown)


benjaminrose:
Is there a way to add a test, or modify a current test like 

https://github.com/sncosmo/sncosmo/blob/865f0c9a5d1a1f15dafdd9a0108e18b03fe33c73/sncosmo/tests/test_salt2utils.py#L110

so that we can make sure this bug is not reintroduced?

ConstanceGanot:
I've modified the test function by adding two 0's at the end of the coefficients list, so that the color law is not modified but the number of coefficients is at the maximum. I tested it on both versions and it raised an error on the previous one.

PR #393, opened 2024-04-08
Fix chisq function returning nan which crashes iminuit

vlad2305m:
This works for me and has not negatively affected the validity of the result of the maximisation.
Please fix the issue of `chisq` returning NaN this way or another

The issue happens because of infinite flux in the model and the invcov matrix is diagonal, so $∞⋅0=$`NaN`

### DIFF
--- a/sncosmo/fitting.py
+++ b/sncosmo/fitting.py
@@ -59,6 +59,8 @@ def generate_chisq(data, model, spectra, signature='iminuit', modelcov=False):
                                             zp=data.zp, zpsys=data.zpsys)
                 diff = data.flux - model_flux
                 phot_chisq = np.dot(np.dot(diff, invcov), diff)
+                if math.isnan(phot_chisq):
+                    phot_chisq = float('inf') # Tell iminuit to keep away from this point (in most cases this should be inf instead of nan anyway)
                 full_chisq += phot_chisq
 
             if spectra is not None:


benjaminrose:
Sorry for letting this sit for so long. But I'm a bit surprised that we are getting Nan at this line. I'm wondering if we want to add a guard against this earlier in the process. Do you have an example of when this happens?

vlad2305m:
As I stated in the first comment, it is enough to just use a prior diagonal invcov matrix and turn a parameter of the model to a high number so that inf appears for flux.
<details>
  <summary>Here's a sample code to showcase it (going up to inf, and then to nan):</summary>
  
```python
import sncosmo
from redback.get_data import get_lasair_data
from sncosmo.fitting import generate_chisq
from sncosmo.photdata import photometric_data
import pandas as pd

model = sncosmo.Model(source='v19-2004et')
model.set(z=0.065)
fparams= ['t0', 'amplitude']

data = get_lasair_data(transient="ZTF23abnmdqn", transient_type='supernova')
sncosmo_data = pd.DataFrame()
sncosmo_data["time"] = data["time"]
sncosmo_data["band"] = data["band"]
sncosmo_data["flux"] = data["flux_density(mjy)"]
sncosmo_data["flux_err"] = data["flux_density_error"]
sncosmo_data["zp"] = [25] * len(data)
sncosmo_data["zpsys"] = data["system"].str.lower()
n_p = len(data)
data = dict(map(lambda col_kv: (col_kv[0], list(map(lambda row_kv: row_kv[1], col_kv[1].items()))), sncosmo_data.items()))

chisq = generate_chisq(photometric_data(data), model, None)
print(model.parameters)
print(data)
print(chisq(0.065, 60244, 1e200)) # This chisq is very very bad (inf).
print(chisq(0.065, 60244, 1e300)) # This chisq is even worse, but it returns nan instead because of how computers work.
```
```
Python 3.11.11
numpy                             1.26.4
pandas                            2.2.2
redback                           1.0.1
sncosmo                           2.10.4
```
</details>

But in reality the function was fitting ok resuslts for χ² and suddenly getting nan for similar reason.
PS. I noticed there's another chisq function with the same problem a few lines below, but it is used by nestle => parameters are bound => it will not overflow, so I didn't notice it.

PR #396, opened 2024-07-17
Remove unused modules

AlexKurek:


benjaminrose:
The failures are from a non-responsive url associated with some built in models.

Issue #372, opened 2023-06-12
Investigate moving dependency from extinction to dust_extinction


benjaminrose:
Do we want to moe to [`dust_extiction`](https://dust-extinction.readthedocs.io/en/stable/index.html)?

benjaminrose:
Ping, @jchiang87

benjaminrose:
all places (not tests or docs) with extinction:

https://github.com/sncosmo/sncosmo/blob/f2a7d6736b48d62788c4dfaf4c7ad68e980adc78/sncosmo/models.py#L1287
https://github.com/sncosmo/sncosmo/blob/f2a7d6736b48d62788c4dfaf4c7ad68e980adc78/sncosmo/models.py#L1997
https://github.com/sncosmo/sncosmo/blob/f2a7d6736b48d62788c4dfaf4c7ad68e980adc78/sncosmo/models.py#L2013
https://github.com/sncosmo/sncosmo/blob/f2a7d6736b48d62788c4dfaf4c7ad68e980adc78/sncosmo/models.py#L2027
https://github.com/sncosmo/sncosmo/blob/f2a7d6736b48d62788c4dfaf4c7ad68e980adc78/sncosmo/models.py#L2032

jvansanten:
Hey, I've just now run into the extinction/numpy v2 issue myself. Has anyone reached out to @kbarbary about transferring the extinction repo to this organization? There's an [open PR that adds numpy2 and py313 support](https://github.com/kbarbary/extinction/pull/27) that only seems to be missing someone to commit to long-term maintenance.

jvansanten:
So, I got curious and compared the implementations of F04, F99, and CCM89 in extinction and dust_extinction.

CCM89 is the same within rounding error, and F99 differs by <0.2%. F04 appears to be something different than extinction's fm07, differing by >5%, even though the comments in dust_extinction seem to imply it was based on the 2007 paper.

Model evaluation is >30x faster in extinction than in dust_extinction, though I'm not sure that's a bottleneck worth worrying about.

Notebook here: https://gist.github.com/jvansanten/952a323b0f0b54acfc84076a2c2c5984

Also, `extinction.apply(mag, flux)` is [literally just `flux * 10**-0.4*mag`](https://github.com/kbarbary/extinction/blob/37b4e3d8e2e9c7f879b7bc32a84b062450542eb5/extinction.pyx#L751-L757)

kbarbary:
[...]

PR #398, opened 2024-09-10
Rebuild for compatibility with Numpy 1.x and 2.x

lpsinger:


benjaminrose:
[extinction](https://github.com/kbarbary/extinction) does not work with numpy 2.0. We could look into moving dependancies, #372.

There may also be other dependancies that are not built for numpy 2.0.

benjaminrose:
I am going to use the same code as extinction:
https://github.com/sncosmo/extinction/blob/8dc8a409ae66d3146505cc6ecae5ceaf85df9750/pyproject.toml#L6-L7

I hope to get this into SNCosmo this week.

PR #414, opened 2026-04-04
arm64 wheels

mcoughlin:
This PR adds 86_64 and arm64 wheels to the build.

### DIFF
--- a/.github/workflows/upload_to_pypi.yml
+++ b/.github/workflows/upload_to_pypi.yml
@@ -13,9 +13,9 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


mcoughlin:
@benjaminrose

mcoughlin:
@benjaminrose thanks. maybe can bump support to 3.10 and above? Not sure if it is worth trying to support python older than that.

benjaminrose:
Yeah. I saw that it was unsupported versions of Python that were failing. I’m not going to bump the supported versions this weekend but I’ll get to it some time next week.

On Apr 4, 2026, at 1:12 PM, Michael Coughlin ***@***.***> wrote:

﻿
[https://avatars.githubusercontent.com/u/2291947?s=20&v=4]mcoughlin left a comment (sncosmo/sncosmo#414)<https://github.com/sncosmo/sncosmo/pull/414#issuecomment-4187519339>

@benjaminrose<https://github.com/benjaminrose> thanks. maybe can bump support to 3.10 and above? Not sure if it is worth trying to support python older than that.

PR #415, opened 2026-05-17
modern python versions

mcoughlin:
This PR updates sncosmo to focus on more modern Python versions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,10 @@ classifiers =
     License :: OSI Approved :: BSD License
     Operating System :: OS Independent
     Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.11
+    Programming Language :: Python :: 3.12
+    Programming Language :: Python :: 3.13
+    Programming Language :: Python :: 3.14
     Topic :: Scientific/Engineering :: Astronomy
     Topic :: Scientific/Engineering :: Physics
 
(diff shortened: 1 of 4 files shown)


mcoughlin:
@benjaminrose

setup.cfg
```diff
@@ -25,22 +29,22 @@ packages =
...
+    astropy>=5.3
     extinction>=0.4.6
     h5py>=3.11
     looseversion>=1.0.0
-    numpy>=1.20.3
+    numpy>=1.23.2
     pyyaml>=6.0.1
     scipy>=1.10.0
-python_requires = >=3.8
+python_requires = >=3.11,<3.15
```
benjaminrose: Why set an upper limit to the python version? We typically don't update these quickly enough to be out on day one of a new python version. I would rather say "we are currently testing on x-versions of python" and let the user try at their own risk on newer versions rather than force a failure.

benjaminrose (commented):
Python 3.10 is still in bug fixes status. Should we remove it now or in the fall when it is end-of-life?

setup.cfg
mcoughlin: I just removed the cap. But I think removing 3.10 makes sense, it is already past schedule to be dropped by the scientific Python community.

benjaminrose (approved):

mcoughlin:
@benjaminrose maybe you can make a release happen?

mcoughlin:
@benjaminrose gentle reminder :)

benjaminrose:
Need to figure out how to keep SNCosmo prioritized. Here is the release on pypi https://pypi.org/project/sncosmo/. Conda coming soon.

PR #418, opened 2026-08-01
Fix band_flux_to_mag to support array input (np.log10 instead of math.log10)

### LINKED ISSUE #403: Using math.log10 on array in MagSystem.band_flux_to_mag can cause error
Hi there,
Thanks for the great package! I think this is a super minor issue. [Line 73 of magsystems.py](https://github.com/sncosmo/sncosmo/blob/v2.10.1/sncosmo/magsystems.py#L71C5-L73C63) currently used `math.log10` which should only take scalars but I assume it was designed to take e.g., np.array as well so maybe should be `np.log10`? 

Thank you again!

ayansark147:
Fixes #403 

## Problem
`MagSystem.band_flux_to_mag` used `math.log10`, which only accepts
Python scalars. Passing a numpy array of flux values (e.g. from a
full light curve) raised:

    TypeError: only 0-dimensional arrays can be converted to Python scalars

## Fix 
Replace `math.log10` with `np.log10` in `sncosmo/magsystems.py`. Since `numpy` was already imported in the file and `np.log10` returns identical results to `math.log10` for scalar input, so this works as replacement with no change in behavior for existing callers.

## Testing
- Existing test suite passes (`pytest sncosmo/tests/ -k magsystem`)
- Manually verified array input now works:

```python
  >>> import sncosmo, numpy as np
  >>> ab = sncosmo.get_magsystem('ab')
  >>> ab.band_flux_to_mag(np.array([1.0, 2.0, 3.0]), 'sdssg')
```


benjaminrose:
Looks good but can you also change/add a test to test the use case of passing a numpy array? Likely this or a similar test, https://github.com/sncosmo/sncosmo/blob/76fa50ca594d47fa37de622c02ab0d1ca5a8b041/sncosmo/tests/test_magsystems.py#L41 I think the documentation is sufficient as is, but it does not explicitly state this capability.

ayansark147:
I added an array input in `test_abmagsystem()`. The pytest passed. Please check and let me know

benjaminrose:
That last test failure is from a download, that sometimes happens. Thanks Ayan for fixing this!

